PR #9, opened 2024-08-20
[oolongwater] iP

oolongwater:
### **GiorgoLite: Your Personal Task Wizard! 🧙‍♂️**

> “I think teeth are the first reminder of mortality. I think that's where it begins to hit ya. Cus they don't come back and that's it.” - Terry A. Davis ([source](https://www.goodreads.com/author/quotes/7171073.Terry_A_Davis))

Giorgo is here to lighten the load on your memory and help you stay on top of your game! It’s:

- Simple to use
- Beautifully minimalistic
- Powerful enough to handle all your tasks effortlessly
- Fast Blazingly fast so you can get things done in a flash ⚡

With Giorgo, you can:

- Type away your tasks in seconds.
- Organize your life like a ~~noob~~ pro.
- Stay focused and ~~stressed~~ stress-free 😊.

Ready to boost your productivity with a touch of magic? 🎩✨

All you need to do is,

- [x] Download the _.jar_ file from [here](https://github.com/oolongwater/ip/releases/tag/A-Jar).
- [x] Copy the _.jar_ file into an empty folder.
- [ ] Open a command window in that folder.
- [ ] Run the command `java -jar "{filename}.jar"` e.g., `java -jar "Duke.jar"` (i.e., run the command in the same folder as the _.jar_ file).

And it is **FREE!**

Features:

1.  Managing tasks
2.  Managing deadlines
3.  Reminders (coming soon)

```java
public class Main {
    }
}
```


src/main/java/MainWindow.java
```diff
@@ -0,0 +1,73 @@
+package main.java;
```
Peanuts359, line 1: I think the GUI logic should be shifted to its own package and kept separate from the core class

src/task/Task.java
```diff
@@ -0,0 +1,85 @@
+package task;
```
Peanuts359, line 1: It's good to see that you moved all your task classes to their own package.

src/ui/Ui.java
```diff
@@ -0,0 +1,132 @@
...
+import task.Task;
+
+
+
+/**
+ * Handles user interactions in the Giorgo application.
+ */
+public class Ui {
+
+    /**
```
Peanuts359, line 18: You actually went to the point of adding JavaDoc comments to EVERY SINGLE METHOD. Nice.

Peanuts359 (commented):
Good job on sticking to even some of the advanced coding standard rules.
There was one minor inconsistency in this regard, but other than that, it's good to go.

src/ui/Ui.java
oolongwater: Thanks!

src/main/java/MainWindow.java
oolongwater: Good idea, thanks

src/storage/Storage.java
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * Loads tasks from the file.
+     *
+     * @return a list of tasks loaded from the file
+     */
+    public List<Task> load() {
+        List<Task> tasks = new ArrayList<>();
+        Path path = Paths.get(filePath);
+
+        try {
```
Reallyeasy1, line 44: hello

Reallyeasy1: May I know why the try catch loop is nested? Is it possible to un-nest the try-catch block?

src/task/Event.java
```diff
@@ -0,0 +1,64 @@
...
+            return null;
+        }
+
+        Event event = null;
+        try {
+            event = new Event(parts[3], parts[4], parts[5], Integer.parseInt(parts[2]));
+        } catch (DateTimeParseException e) {
+            e.printStackTrace();
+        }
+        assert event != null;
```
Reallyeasy1, line 60: In the try catch block, if there is DateTimeParseException, event can be null.
The assertion line may raise an exception. Are there any ways to handle this?

Reallyeasy1, line 60: In the try catch block, if there is DateTimeParseException, event can be null. The assertion line may raise an exception. Are there any ways to handle this?

src/ui/Ui.java
```diff
@@ -0,0 +1,184 @@
...
+        } catch (DateTimeParseException e) {
+            // If parsing fails, treat the argument as a normal string
+        }
+
+        for (Task task : tasks) {
+            if (inputDate != null) {
+                if (task instanceof Deadline deadline) {
+                    if (deadline.getBy().toLocalDate().isEqual(inputDate)) {
+                        sb.append(count).append(") [Priority: ").append(
+                                task.getPriority()).append("] ").append(task).append("\n\n");
```
Reallyeasy1, line 119: Are there any ways of un-nesting the if-else nests?

src/ui/Ui.java
```diff
@@ -0,0 +1,184 @@
...
+                        count++;
+                    }
+                } else if (task instanceof Event event) {
+                    LocalDate startDate = event.getFrom().toLocalDate();
+                    LocalDate endDate = event.getTo().toLocalDate();
+                    if ((inputDate.isEqual(startDate) || inputDate.isAfter(startDate))
+                            && (inputDate.isEqual(endDate) || inputDate.isBefore(endDate))) {
```
Reallyeasy1, line 126: Is it possible to split this into individual boolean values?
(eg. isGTEStartDate and isLTEEndDate)

src/ui/Ui.java
```diff
@@ -0,0 +1,184 @@
...
+        StringBuilder sb = new StringBuilder("Here are the tasks on " + date + ":\n\n");
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd");
+        LocalDate inputDate = LocalDate.parse(date, formatter);
+        int count = 1;
+
+        for (Task task : tasks) {
+            if (task instanceof Deadline deadline) {
+                if (deadline.getBy().toLocalDate().isEqual(inputDate)) {
+                    sb.append(count).append(") [Priority: ").append(
+                            task.getPriority()).append("] ").append(task).append("\n\n");
```
Reallyeasy1, line 163: Are there any ways of un-nesting the if-else nests?

src/ui/Ui.java
```diff
@@ -0,0 +1,184 @@
...
+                    count++;
+                }
+            } else if (task instanceof Event event) {
+                LocalDate startDate = event.getFrom().toLocalDate();
+                LocalDate endDate = event.getTo().toLocalDate();
+                if ((inputDate.isEqual(startDate) || inputDate.isAfter(startDate))
```
Reallyeasy1, line 169: Are there any ways of un-nesting the if-else nests?

src/ui/Ui.java
```diff
@@ -0,0 +1,184 @@
...
+                }
+            }
+        }
+
+        if (count == 1) {
+            return "No tasks found on " + date;
+        }
+
+        return sb.toString();
+    }
```
Reallyeasy1, line 183: Good job in writing your code, relatively minor changes only.

PR #15, opened 2024-08-20
[Wong Zheng Jie] iP

KrashKart:
# Jackson
> At any one time, language is a kaleidoscope of genres, styles and dialects - [David Crystal](https://quotefancy.com/quote/1151562/David-Crystal-At-any-one-time-language-is-a-kaleidoscope-of-styles-genres-and-dialects)

Jackson is not :no_entry_sign: *any* ordinary chat-bot and task tracker. Devoid of posh and proper English, Jackson embraces the imperfection of **Singlish** in a uniquely ~~rabakz~~ Singaporean way :metal:.

## Installing Jackson
Before you *lim kopi* :coffee: (drink coffee) with Jackson, ensure that you:
1. Download the jar file [here](https://github.com/KrashKart/ip/releases/latest)
2. Run the jar file
3. Start adding your tasks!

## Why Jackson?
- Trust Jackson to track your tasks from events to deadlines :clipboard: 
- Simply `mark` or `unmark` your tasks to track completion :white_check_mark:
- Forgot a specific task? Jackson comes with in-built `find` functionalities :mag:
- Up for the challenge? We dare you to find a ~~cool easter egg~~ within Jackson :shushing_face:

## Up and Coming Features
- [x] Add save file saving and reading features for continuity
- [x] Add standardised date formatting and storage
- [x] Add GUI
- [x] Update easter egg

## Exploring with Jackson
If you would like to experiment or integrate Jackson into your own products, provided below is the `main` code:
```java
    // create instance of Jackson and run main loop
    Jackson jackson = new Jackson(EXPECTED_SIZE, PATH);
    jackson.run();
}
```
where `EXPECTED_SIZE` and `PATH` are the expected task list sizes (can be violated dynamically) and the path for the save file (can be set dynamically).


src/main/java/jackson/utils/Parser.java
```diff
@@ -0,0 +1,106 @@
...
+
+import jackson.enums.Actions;
+import jackson.exceptions.SyntaxException;
+
+/**
+ * Parser static class to handle regex checking and input parsing.
+ */
+public class Parser {
+
+    /* Bunch of regex patterns for user input checking */
```
btbrandon: Perhaps you can elaborate more on what regex does?

src/main/java/jackson/Jackson.java
```diff
@@ -0,0 +1,225 @@
...
+            output = secret;
+        }
+        // print secret msg!
+        return output;
+    }
+
+    /**
+     * Runs the main loop of the chatbot.
+     */
+    public String getResponse(String input) {
```
btbrandon, line 78: code looks very structured and well-designed!

src/main/java/jackson/exceptions/UnsupportedException.java
```diff
@@ -0,0 +1,10 @@
+package jackson.exceptions;
+
+/**
+ * Thrown when unrecognised command and input is given.
+ */
+public class UnsupportedException extends JacksonException {
```
btbrandon: could have a clearer name for describing what exactly is unsupported

src/main/java/jackson/utils/Parser.java
```diff
@@ -0,0 +1,106 @@
...
+     * @param  query String containing user input.
+     * @return Response containing Action Type and regex Matcher object.
+     * @throws SyntaxException when unknown query is parsed.
+     */
+    public static Response parse(String query) throws SyntaxException {
+        Actions.ActionType a;
+        Matcher m = null;
+
+        /*
+            Generally:
```
btbrandon, line 46: nice comment! makes it clearer

src/main/java/jackson/utils/Temporal.java
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Class used to handle LocalDate and LocalDateTime objects.
+ * This is required as we want to allow flexibility in whether the user defines
+ * a time or not, and both LocalDate and LocalDateTime strictly require no timing
+ * and timing respectively.
+ */
+public class Temporal {
+    // DateTimeFormatters for reading (first 2) from save file and user input
+    // and writing (last 2) to save file
+    private static final DateTimeFormatter dateOnly = new DateTimeFormatterBuilder()
```
btbrandon: naming conventions are specific and clearly differentiates the meaning good job!

src/main/java/jackson/Jackson.java
```diff
@@ -0,0 +1,225 @@
...
+            try {
+                // read via scanner line by line
+                Scanner sc = new Scanner(f);
+                while (sc.hasNextLine()) {
+                    temp.append(sc.nextLine()).append("\n");
+                }
+                output = temp.toString().strip();
+
+                // decode from base64 to utf8
+                byte[] decoded = Base64.getDecoder().decode(output);
```
vicelyav: Maybe this one could add more description to the variable on what is being decoded?

src/main/java/jackson/Jackson.java
```diff
@@ -0,0 +1,225 @@
...
+        this.storage = new Storage(PATH);
+        this.commandType = Commands.CommandType.INTRO;
+    }
+
+    /**
+     * Reads secret text from secret file and prints it.
+     * If secret file not found, handles exception and prints error message.
+     * @return String response.
+     */
+    public String readSecret() {
```
vicelyav: Maybe we could rename it as "readSecretFile()" so we know what is secret.

src/main/java/jackson/Jackson.java
```diff
@@ -0,0 +1,225 @@
...
+        // get string builder to read line by line
+        StringBuilder temp = new StringBuilder();
```
vicelyav: Maybe could rename this variable a bit to show what temp String we're using it for?

src/main/java/jackson/utils/Ui.java
```diff
@@ -0,0 +1,213 @@
...
+        return output;
+    }
+
+    /**
+     * Returns chatbot response when deleting task from list.
+     * @param task {@code Task} object to be deleted from taskList.
+     * @param taskList {@code TaskList} object task is to be deleted from.
+     * @return String response.
+     */
+    public String printDeleteList(Task task, TaskList taskList) {
```
vicelyav: Maybe we could rename this to show that we're printing after deleting something on the list? Right now when I first read it, I thought it is print, then delete list.

src/main/java/jackson/utils/Ui.java
```diff
@@ -0,0 +1,213 @@
...
+        }
+        return output;
+    }
+
+    /**
+     * Returns chatbot response when marking task in list.
+     * @param task {@code Task} object to be marked.
+     * @return String response.
+     */
+    public String printMark(Task task) {
```
vicelyav: Maybe we could rename this to show that we are printing Marked Task instead of printing Mark (could be misinterpreted)?

vicelyav (commented):
Overall good job on writing a very neat code. I think for the methods, some of them can be renamed/edited to show clearer what the method is doing. Cheers!

src/main/java/jackson/Jackson.java
KrashKart: thanks for the review! however, i wanted to make this an easter egg (albeit poorly hidden) as much as possible, so had to be vague with the names here

src/main/java/jackson/utils/Ui.java
KrashKart: fair point! what i had in mind was that the command given was "deleteList", thus to print that command the method should be called "printDeleteList", but i get your point. will change it and thanks for the review :+1:

src/main/java/jackson/utils/Parser.java
KrashKart: fair enough! will update the code to explain a bit more about regex. thanks for the review!

PR #24, opened 2024-08-21
[wkxcass] iP

wkxcass:
### MoiMoi

> "Your conscious mind, like the computer screen, is a focusing tool, not a storage place." – David Allen ([source](https://earnworthy.com/david-allen-quote/))

MoiMoi helps you _focus on what matters_, freeing up your mind from the clutter of tasks.

MoiMoi is an **easy-for-all** task management solution. It's,
- text-based 📝
- user-friendly 🤝
- ~fast~ _SUPER_ FAST to use ⚡

MoiMoi's features include,
- [ ] Task management
- [ ] Deadline tracking (coming soon)
- [ ] Reminder notifications (coming soon)

And, the best part – it's for **FREE**! 🎉

Getting MoiMoi is as easy as ABC:
1. Ensure that `Java 17` is installed in your computer.
2. Download the `.jar` file of the latest release, from [here](https://github.com/wkxcass/ip/releases/).
3. Move the `.jar` file into an empty folder.
4. Open a command window in that folder (right-click on the folder, then select `Open in Terminal`).
5. Run the command `java -jar "moimoi.jar"`.

If you are a Java programmer, you can use MoiMoi's source code to practice Java too! Here's the `main` method:
```java
public class Main {
    }
}
```


src/main/java/moimoi/gui/DialogBox.java
```diff
@@ -0,0 +1,82 @@
...
+
+        this.dialog.setText(text);
+        this.displayPicture.setImage(img);
+    }
+
+    /**
+     * Flips the dialog box such that the ImageView is on the left and text on the right.
+     */
+    private void flip() {
+        ObservableList<Node> tmp = FXCollections.observableArrayList(this.getChildren());
```
Teddayz, line 52: Perhaps could be named as tempList instead to give yourself more information what the variable is :)

src/main/java/moimoi/gui/DialogBox.java
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Returns a dialog box with the specified image of and text from MoiMoi.
+     *
+     * @param text Text from MoiMoi.
+     * @param img Image of MoiMoi.
+     * @return Dialog box with the image of and text from MoiMoi.
+     */
+    public static DialogBox getMoiMoiDialog(String text, Image img) {
+        var db = new DialogBox(text, img);
```
Teddayz, line 77: Likewise over here, can give a better name for yourself to know what it is. Can be var dialogBox

src/main/java/moimoi/gui/MainWindow.java
```diff
@@ -0,0 +1,90 @@
...
+                DialogBox.getMoiMoiDialog(this.moiMoi.getInitialMessage(), this.moiMoiImage));
+    }
+
+    /**
+     * Initializes the scroll pane,
+     * by binding its vertical scroll value to the height of the dialog container.
+     */
+    @FXML
+    public void initialize() {
+        scrollPane.vvalueProperty().bind(dialogContainer.heightProperty());
```
Teddayz, line 66: vValueProperty would be better

src/main/java/moimoi/util/exception/InvalidDateTimeException.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    /**
+     * Returns corresponding error message.
+     *
+     * @return Error message.
+     */
+    @Override
+    public String getMessage() {
+        return super.getMessage() + "The " + this.type
```
Teddayz, line 36: Maybe using String.format will reduce the length of this statement so that its more readable

src/main/java/moimoi/util/task/Deadline.java
```diff
@@ -0,0 +1,62 @@
...
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Represents a deadline task.
+ */
+public class Deadline extends Task {
+
+    protected LocalDateTime by;
```
Teddayz: Perhaps the name could be better, for example maybe we can use deadline as the name so that we know what the variable is :)

Teddayz (commented):
Overall, LGTM, has good coding style practices. 
One observation is that Deadline Tasks, event tasks etc can be renamed to DeadlineTask, EventTask, ToDoTask etc as it can make it more understandable that it is inherited from Task. 👍 Overall, good job!

src/main/java/moimoi/util/task/Task.java
```diff
@@ -0,0 +1,91 @@
...
+     */
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); // mark done task with X
+    }
+
+    /**
+     * Marks the task as done.
+     */
+    public void mark() {
+        this.isDone = true;
```
JunLongling, line 39: Hey, I noticed that this.isDone is used here, but since the field isn't shadowed by a parameter, I'm wondering if it's necessary. Should it be simplified to `isDone = true;` to reduce noise? I saw a similar pattern in a few other places as well—perhaps it's worth reviewing those instances too. Let me know what you think!

src/test/java/moimoi/util/command/AddCommandTest.java
```diff
@@ -0,0 +1,169 @@
...
+import moimoi.util.exception.MoiMoiException;
+import moimoi.util.task.Task;
+
+public class AddCommandTest {
+
+    private Storage storage = new Storage("data/test.txt");
+    private TaskList tasks = new TaskList(new ArrayList<Task>());
+
+    @Test
+    public void testisExit() {
```
JunLongling: Would `testIsExit` be the correct naming standard here?

src/main/java/moimoi/util/task/Deadline.java
```diff
@@ -0,0 +1,62 @@
...
+     * @return true if the task occurs on the specified date; false otherwise.
+     */
+    @Override
+    public boolean occursOn(LocalDate date) {
+        if (this.by.toLocalDate().isEqual(date)) {
+            return true;
+        } else {
+            return false;
+        }
+    }
```
JunLongling: Great job with the occursOn method! I see you’ve implemented the date comparison effectively using isEqual. While your approach works perfectly, you could consider simplifying the method by returning the result of the condition directly, like this: `return this.by.toLocalDate().isEqual(date);` This way, the logic remains the same but the code becomes more concise. Of course, there’s no right or wrong here—just a suggestion for clarity. What do you think?

src/main/java/moimoi/util/task/Task.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    /**
+     * Marks the task as undone (i.e., unmarks the task).
+     */
+    public void unmark() {
+        this.isDone = false;
```
JunLongling, line 46: Would it be better to simplify this to` isDone = false`; to reduce noise?

JunLongling (commented):
Overall, the coding standards are mostly adhered to (or at least, I can't detect many violations), good job! Keep up the great work! 👍

src/main/java/moimoi/gui/MainWindow.java
wkxcass: Thank you for noticing! However there's nothing much I can do to it, since `vvalueProperty` is the name of an instance method of the `ScrollPane` class 😆

src/main/java/moimoi/util/task/Deadline.java
wkxcass: Thanks! I agree with your feedback👍 I'll tweak my code accordingly to improve variable readability.

src/main/java/moimoi/util/task/Task.java
wkxcass: Thank you so much for the insightful feedback! I hadn’t considered that perspective before.
My original intent behind using `this` was to remind myself that the variables are class instance scoped. However, you do make a point about reducing unnecessary noise, as `this` isn’t needed in most cases throughout the code. I’ll give this further consideration—it's indeed something worth thinking about, though I’m not ready to commit to refactoring just yet~

src/test/java/moimoi/util/command/AddCommandTest.java
wkxcass: Yes, seems like I overlooked this part 😆 Thanks for pointing it out!

src/main/java/moimoi/util/task/Deadline.java
wkxcass: I like your suggestion! I'll tweak my code accordingly to make it more concise. Thanks for bringing up! 😄

PR #34, opened 2024-08-22
[hm-yang1] iP

hm-yang1:
# CheeseBot
>"Life is great. Cheese makes it better." -- Anonymous

CheeseBot frees your mind of having to remember things you need to do. It's:
- text-based
- easy to learn
- ~FAST~ SUPER FAST to use

All you need to do is:
- download it from [here](https://github.com/hm-yang1/ip).
- double-click it.
- add your tasks.
- let it manage your tasks for you 😄 

And it is **FREE!**

Features:
- [x] Managing tasks
- [ ] Managing deadlines
- [ ] Reminders
---
```
public class WheelyBigCheese {
        WheelyBigCheese bot = new WheelyBigCheese();
        bot.run();
    }
}
```


src/main/java/cheese/CheeseException.java
```diff
@@ -0,0 +1,14 @@
+package cheese;
+
+/**
+ * Custom exception
```
myrilla12: JavaDoc header comments should start in the form of something like 'Customizes...' or 'Represents...'

src/main/java/cheese/CheeseException.java
```diff
@@ -0,0 +1,14 @@
...
+ */
+public class CheeseException extends Exception {
+    /**
+     * Just a constructor
```
myrilla12: This is a very minor comment but I think there should be a blank space in between the descriptor and any @param or @return. Each line should also be concluded with a full stop.

src/main/java/cheese/Parser.java
```diff
@@ -0,0 +1,81 @@
...
+ * Parses user input
+ */
+public class Parser {
+    /**
+     * Get idx of item in list from user input
+     * @param inputTokens input from user
+     * @return int
+     * @throws CheeseException Missing/Incorrect input
+     */
+    public static int getIdx(String[] inputTokens, int size) throws CheeseException {
```
myrilla12: Could idx be a potentially confusing name? Perhaps consider extending it to index.

src/main/java/cheese/Storage.java
```diff
@@ -0,0 +1,106 @@
...
+        case "E" -> new Event(data);
+        default -> throw new CheeseException("Incorrect data format");
+        };
+    }
+
+    /**
+     * Add and save new task
+     * @param t task to be added
+     */
+    public void add(Task t) throws CheeseException {
```
myrilla12: Could the naming appear a little inconsistent? The previous method was called loadTask while the following are just add and update rather than addTask and updateTask

src/main/java/cheese/TaskList.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+    public TaskList(ArrayList<Task> tasks) {
+        this.tasks = tasks;
+    }
+
+    public void add(Task task) {
+        tasks.add(task);
+    }
+
+    public int size() {
```
myrilla12, line 26: Shouldn't the method name be a verb and not a noun? For example, getSize instead of size

src/main/java/cheese/TaskList.java
```diff
@@ -0,0 +1,36 @@
...
+        return tasks.size();
+    }
+
+    public Task get(int i) {
```
myrilla12, line 30: I personally find 'get' to be a vague method name

src/main/java/cheese/command/AddCommand.java
```diff
@@ -0,0 +1,37 @@
...
+import cheese.Storage;
+import cheese.TaskList;
+import cheese.Ui;
+import cheese.task.Task;
+
+/**
+ * Command to add tasks to list
+ */
+public class AddCommand extends Command {
+    private final Task task;
```
myrilla12, line 13: Shouldn't final variables be named in SCREAMING_SNAKE_CASE?

src/main/java/cheese/command/Command.java
```diff
@@ -0,0 +1,36 @@
...
+import cheese.CheeseException;
+import cheese.Storage;
+import cheese.TaskList;
+import cheese.Ui;
+
+/**
+ * Encapsulates small parts of logic for the bot
+ */
+public class Command {
+    private final boolean exitChat;
```
myrilla12, line 12: Same with this final variable

myrilla12 (commented):
LGTM. I think the JavaDoc comments might need a look over as they don't follow comment conventions. However, the code mostly follows convention aside from a few naming conventions.

src/main/java/cheese/command/Command.java
ty4g1, line 12: To follow the coding standard, perhaps it would be better to rename this boolean to `isExitChat` or `isExitCommand` instead.

src/main/java/cheese/command/MarkCommand.java
```diff
@@ -0,0 +1,40 @@
...
+import cheese.Storage;
+import cheese.TaskList;
+import cheese.Ui;
+import cheese.task.Task;
+
+/**
+ * Command to mark a task as done
+ */
+public class MarkCommand extends UpdateCommand {
+    private final boolean done;
```
ty4g1: Similar comment, would be better to rename it to `isDone` to follow the coding standard.

ty4g1 (commented):
LGTM, just a few nits with variable names to follow coding standard. I agree with the other comments, the JavaDocs could be more descriptive.

PR #63, opened 2024-08-23
[Kelly Wang] iP

kellywsq03:
## ECHO: Your one-stop planner 😱

> Echo changes your **to-dos** to _to-dones_ ;)

Become a [productivity pro](https://ca.indeed.com/career-advice/career-development/daily-to-do-list)!

Why use to-do GUIs that looks like this:
- [x] Walk dog
- [ ] Wash clothes
- [ ] Go for a run

When you can CHAT with ECHO!!
Chat with Echo to plan tasks as a:
- Todo
- Event
- Deadline

Just follow these steps:
1. Download the build
2. Run the build with gradle
3. Have fun!!

Take a sneak peak at our code, here's the `main` method:
```/**
     * The main method that launches the Echo application.
     *
     * @param args Command-line arguments (not used).
     */
        new Echo("src/main/data/savedTasks.txt").run();
    }

### CODE REVIEW

**Review by Brendan8899 — COMMENTED** [review]
Good job defining your commands as Enumerations as commands belong to a fixed set. The capitalisation of command names is appropriate as constant names must all be uppercase

File: src/main/java/echo/Echo.java [hunk 1]
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    /**
+     * Generates a response for the user's chat message.
+     */
+    public String getResponse(String input) {
+        return ui.handleInput(input);
+    }
+
+    public void stopRunning() {
```
LOLIPOP-INTELLIGENCE, line 43: Missing header comments

src/main/java/echo/Echo.java
```diff
@@ -0,0 +1,44 @@
...
+        storage.saveToFile();
+    }
+
+    public String greetUser() {
```
LOLIPOP-INTELLIGENCE, line 54: Missing header comments.

src/main/java/echo/Ui.java
```diff
@@ -0,0 +1,295 @@
+package echo;
```
LOLIPOP-INTELLIGENCE, line 1: The import statements should be reordered according to the coding standard. They should be grouped as follows: static imports (if any), Java imports, third-party imports (if any), then project imports.

src/main/java/echo/backend/Parser.java
```diff
@@ -0,0 +1,187 @@
...
+     * @throws DateTimeParseException if the date string cannot be parsed with any of the formats
+     */
+    public LocalDate parseDate(String s) throws DateTimeParseException {
+        DateTimeFormatter formatter;
+        for (String pattern : DATE_FORMATS) {
+            formatter = DateTimeFormatter.ofPattern(pattern);
+            try {
+                LocalDate localDate = LocalDate.parse(s, formatter);
+                return localDate;
+            } catch (DateTimeParseException e) {
```
LOLIPOP-INTELLIGENCE, line 241: Consider adding a comment why the handling of this exception is ignored.

PR #70, opened 2024-08-24
[sirui] iP

beginner-web-developer:
### Alice Chatbot

> “A computer is like a violin. You can imagine it making beautiful music, but you have to learn how to play it.” -- Bill Gates ([source](https://goglobalways.com/blog/inspiring-computer-science-quotes/))

It is a CLI application, as such it is **SUPER FAST** to use! Being a CLI application, which only accepts specific comments currently, it is important to know which commands to input to get the right response. Current commands include:
- "list"
- "bye"
- "todo"
- "deadline"
- "event"
- "mark"
- "unmark"
- "delete"

Features:
- [x] Managing tasks
- [ ] Managing deadlines (_coming soon_)
- [ ] Reminders (_coming soon_)

All you need to do is,
1. Download it from [here](https://nus-cs2103-ay2425s1.github.io/website/schedule/week4/project.html).
2. Double-click it.
3. Add your tasks.
4. Let it manage your tasks for you 😉

The code uses OOP. Here is an example snapshot of the `Alice` class using other classes such as `Scanner`, `Storage` and `Ui`.
```java
public class Alice {
    private final TaskList TASKS;
    private final Storage STORAGE;
    private final Ui UI;
}
```


src/main/java/Alice.java
```diff
@@ -0,0 +1,44 @@
...
+    public Alice(String filePath) {
+        storage = new Storage(filePath);
+        // load tasks
+        storage.loadTasks();
+        tasks = storage.getTasks();
+
+        // initialise UI with the loaded tasks
+        ui = new Ui(tasks);
+    }
+
```
jinnsuke, line 30: I like how you've structured the initialization of Storage, TaskList, and Ui in the constructor—it makes the flow of the setup quite clear. I’m wondering, though, would it be beneficial to extract the task-loading logic into a separate method? It could make the constructor more concise and improve readability. Just a thought for making the code a bit cleaner!

src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,136 @@
...
+            break;
+        case DELETE:
+            int taskNo;
+            try {
+                taskNo = Integer.parseInt(result[1]);
+            } catch (NumberFormatException e) {
+                System.out.println(result[1] + " is not a number!");
+                System.out.println("------------------------------------------");
+                break;
+            }
```
jinnsuke: I like how you've implemented error handling for invalid task numbers with a try-catch block! It ensures that the program doesn't crash when a user inputs a non-numeric value. However, would it be more user-friendly to handle the error by providing a more detailed message or even prompting the user to re-enter a valid number? This could improve the user experience by guiding them on how to correct their input. What do you think about this approach?

src/main/java/storage/TaskList.java
```diff
@@ -0,0 +1,158 @@
...
+            String[] times = response[1].split("/to ");
+            task = new Event(response[0], times[0].strip(), times[1].strip());
+            break;
+        default:
+            break;
+        }
+        tasks.add(task);
+        return task;
+    }
+
```
jinnsuke, line 75: The addTask method in the TaskList class handles both task creation and addition. While this approach works, it might be worth considering separating the task creation logic into a dedicated method or class. This would enhance the Single Responsibility Principle (SRP), making the code easier to maintain, especially if task creation logic becomes more complex in the future. This separation could also make the method cleaner and more focused on adding tasks to the list. What do you think about this?

src/main/java/ui/Ui.java
```diff
@@ -0,0 +1,139 @@
...
+        showDivider();
+    }
+
+    /**
+     * Prints the divider to the terminal.
+     */
+    public void showDivider() {
+        System.out.println(Ui.DIVIDER);
+    }
+}
```
jinnsuke, line 113: The Ui class is well-structured and easy to read, with each method having a clear and focused responsibility. The separation of concerns is well-executed, making the code intuitive to understand and maintain. I particularly appreciate how methods like printHandleTaskMessage and showDivider are used to encapsulate specific actions, ensuring that the code remains organised and avoiding repetitive code. This design enhances the overall readability and makes it easy to extend or modify the class in the future. Great job!

jinnsuke (commented):
LGTM. Just some simple recommendations!

src/main/java/DialogBox.java
```diff
@@ -0,0 +1,58 @@
...
+
+        dialog.setText(text);
+        displayPicture.setImage(img);
+    }
+
+    /**
+     * Flips the dialog box such that the ImageView is on the left and text on the right.
+     */
+    private void flip() {
+        ObservableList<Node> tmp = FXCollections.observableArrayList(this.getChildren());
```
sayomaki, line 43: I know this naming could be better but the javadoc does help explain what the method does!

sayomaki, line 43: I know this naming could be better but the javadoc does help explain what the method does!

src/main/java/MainWindow.java
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Controller for the main GUI.
+ */
+public class MainWindow extends AnchorPane {
+    @FXML
+    private ScrollPane scrollPane;
+    @FXML
+    private VBox dialogContainer;
+    @FXML
+    private TextField input;
```
sayomaki, line 21: For this name, it may be not so obvious what this input is referring to (until you find where the input is used). A better name could be `userInput`

src/main/java/exceptions/AliceException.java
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    /**
+     * Returns the string representation of the exception.
+     *
+     * @return custom error message for AliceException.
+     */
+    @Override
+    public String toString() {
+        return "OOPS! I don't know what '" + this.message + "' means!";
```
sayomaki, line 21: `this` keyword may be omitted as there is no overlapping variable with the same name.

src/main/java/ui/Ui.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    /**
+     * Returns a string of all the tasks in the list.
+     *
+     * @param tasks the string representation of all the tasks.
+     * @param size number of tasks in the list.
+     * @return tasks in the list.
+     */
+    public String getListedTasks(String tasks, int size) {
```
sayomaki: Maybe for this method, you can follow the naming for the other methods in the class (ending with the -Message suffix), I think this would be helpful for others to understand and to being consistent!

sayomaki (commented):
Good to see that the code is mostly following good code naming practices and the methods are very concise, but with just a few little comments as added below.

Code is also well structured and easy to understand and follow!

PR #71, opened 2024-08-24
[liauzhanyi] iP

matchastack:
# ZBot

This PR adds **more functionalities** to ~~Duke~~ ZBot. 

The bot can now keep track of tasks, and you may even specify due dates, or a time period just like a calendar 📆 event!

## User Guide

To use the app, follow the steps below:

1. Clone the [repository](https://github.com/liauzhanyi/ip)
2. Run the program using 
```bash
./gradlew build
./gradlew run
```

## Command Summary

These are the currently supported commands

- `todo [task]`
- `deadline [task] /by [dd/MM/yyyy HHmm]`
- `event [task] /from [dd/MM/yyyy HHmm] /to [dd/MM/yyyy HHmm]`
- `list`
- `delete [index]`
- `find`
- `mark [index]` 
- `unmark [index]`

More features _coming soon_!

- [ ] GUI
- [ ] Support for more date input formats

For future developers, the aim of the project is to be
> everything to someone

and not 

> something to everyone

Since there is no syntax highlight for bash, this is just here for the sake of adding colours since as programmers, we all love _colourful words_

```java
import java.utils.*;

    new ZBot(SAVE_PATH).run();
}
```


src/main/java/zbot/Storage.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    /**
+     * Writes text to the file.
+     *
+     * Overwrites the file if it already exists.
+     *
+     * @param text Text to write to the file.
+     */
+    public void writeToFile(String text) {
```
df-Nic: A suggestion will be to indicate the type of file to be written into in the method name.

**For example:**
```java
public void writeToTextFile(String text) {
    // Function Body
}
```

src/main/java/zbot/task/Event.java
```diff
@@ -0,0 +1,50 @@
...
+ * Represents an event with a description, start and end datetime.
+ */
+public class Event extends Task {
+    protected LocalDateTime startDate;
+    protected LocalDateTime endDate;
+
+    /**
+     * Constructor for Event.
+     *
+     * @param description Description of the event.
```
df-Nic, line 17: If I am not mistaken, for JavaDocs there is not need to align the description for each entry?

src/main/java/zbot/ui/MainWindow.java
```diff
@@ -0,0 +1,58 @@
+package zbot.ui;
+
+import javafx.fxml.FXML;
+import javafx.scene.control.Button;
+import javafx.scene.control.ScrollPane;
+import javafx.scene.control.TextField;
+import javafx.scene.image.Image;
+import javafx.scene.layout.AnchorPane;
+import javafx.scene.layout.VBox;
+import zbot.ZBot;
```
df-Nic, line 10: Possibly add a space between the imports of different packages.

**For example:**
```java
//...
import javafx.scene.layout.VBox;

import zbot.Zbot
```

To follow the standardization of the other imports in the other files, which is good that you have a standard for the other java files as well!

src/main/java/zbot/Parser.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+/**
+ * Represents a parser that parses and formats date and time.
+ */
+public class Parser {
+    private static final String DEFAULT_TIME = "2359";
```
df-Nic, line 11: Good that you follow the snake case naming convention for static variables.

src/main/java/zbot/Storage.java
```diff
@@ -0,0 +1,141 @@
...
+     *
+     * @param filePath Path of the file to store tasks.
+     */
+    public Storage(String filePath) {
+        this.filePath = filePath;
+    }
+
+    /**
+     * Creates the file if it does not exist.
+     *
```
df-Nic: A suggestion to add <p> to denote a new paragraph.

If I am not mistaken, this will render the description as 1 line. Therefore possibly try this.

**For example:**
```java
    /**
     * Creates the file if it does not exist.
     * <p>
     * If the file does not exist, the folder containing the file will be created as
     * well.
     * <p>
     * If the file already exists, nothing will be done.
     */
    public void createFileIfNotExists() {
        // Function Body
    }
```

src/main/java/zbot/ZBot.java
```diff
@@ -0,0 +1,228 @@
...
+     *
+     * The input is processed and the corresponding action is taken.
+     *
+     * @param input User input.
+     * @return Response to the user input.
+     */
+    public String executeInput(String input) {
+        String response;
+
+        if (input.equals("list")) {
```
df-Nic, line 73: Good that you follow the indentation format for the `if`, `else` statements 👍

df-Nic (commented):
Overall the quality of the code follows the standard shown in the lecture materials 💯

src/main/java/zbot/Storage.java
matchastack: makes sense, I will make the change. thanks!

src/main/java/zbot/task/Event.java
matchastack: sorry I do not understand what is meant by "align the description for each entry" can you provide some examples? Thank you.

src/main/java/zbot/ui/MainWindow.java
matchastack: yes! I plan to do that but it will not pass the `checkstyle.xml` from AB3. I am still finding out how to edit the file. Thanks for pointing it out though!

src/main/java/zbot/Storage.java
matchastack: I see... did not know that. Thanks!

src/main/java/zbot/task/Event.java
df-Nic: Sorry for the late reply

```
/**
 * Constructor for Event.
 *
 * @param description Description of the event.
 * @param from        Start datetime of the event.
 * @param to          End datetime of the event.
 */
```
From what I see the JavaDoc looks like this.

I just mean that for the other 2 params, `from` and `to` there is no need to have the extra white spaces in-between the variable name and the description of the variable, if I recall correctly.

matchastack: I see what you mean...

Yes, thanks for pointing it out. I agree it's not required but I don't think it violates the standard too as long as it is kept consistent across. 

I will take a look across the other files and choose a format for consistency. Thanks!

PR #83, opened 2024-08-24
[sooyitao] iP

sooyitao:
# HanaBot
> “Productivity is never an accident. It is always the result of a commitment to excellence, intelligent planning, and focused effort.” – [Paul J. Meyer](https://en.wikipedia.org/wiki/Paul_J._Meyer)

**HanaBot** is a chatbot that helps you organize your tasks efficiently. It's:
- Simple and intuitive to use
- Lightweight and ~~fast~~ _SUPER FAST_
- Command-line based for simplicity

### Getting Started
1. [Download HanaBot .jar file](https://github.com/sooyitao/ip/releases/tag/A-Jar)
2. `cd` into the folder of the downloaded jar file
3. Run `java -jar Hana.jar`
4. Start adding tasks to your task list! 🫡
5. Let it manage your tasks for you 😉

### What's New
- [x] Task Management: Add, edit, and remove tasks.
- [x] Deadlines: Set deadlines for your tasks.
- [ ] Recurring Tasks: Automate repetitive tasks (coming soon).
- [ ] Notifications: Receive reminders for upcoming tasks (coming soon).

**HanaBot** is also a great tool for Java developers to sharpen their skills. Here’s a snippet of the main entry point for the application:
``` java
public class Hana {
        new Hana().run();
    }
}
```


README.md
Nimastic, line 1: Looks good to me 👍

Nimastic (commented):
Looks good to me 😄

src/main/java/hana/command/DeadlineCommand.java
```diff
@@ -0,0 +1,58 @@
...
+     * Constructs a new DeadlineCommand with a description and due date.
+     *
+     * @param input The input from user.
+     * @throws HanaException If an error occurs.
+     */
+    public DeadlineCommand(String input) throws HanaException {
+        try {
+            String[] parts = input.substring(8).split(" /by ");
+            if (parts.length < 2) {
+                throw new HanaException("Deadline task must have a description and a due date.");
```
tohhengxing, line 32: nit: can be a constant

src/main/java/hana/command/EventCommand.java
```diff
@@ -0,0 +1,60 @@
...
+     */
+    public EventCommand(String input) throws HanaException {
+        try {
+            String[] parts = input.substring(5).split(" /from | /to ");
+            if (parts.length < 3) {
+                throw new HanaException("Event task must have a description, start time, and end time.");
+            }
+            DateTimeFormatter formatter = DateTimeFormatter.ofPattern("d/M/yyyy HHmm");
+            this.from = LocalDateTime.parse(parts[1].trim(), formatter);
+            this.to = LocalDateTime.parse(parts[2].trim(), formatter);
```
kuiktzecheng, line 37: can consider to have 2 constants : INPUT_FORMATTER and OUTPUT_FORMATTER to format the input date and times

src/main/java/hana/ui/Ui.java
```diff
@@ -0,0 +1,138 @@
...
+     * @param message Message to print.
+     */
+    public void printMessage(String message) {
+        responseBuilder.append(message).append("\n");
+    }
+
+    /**
+     * Prints all commands.
+     */
+    public void printCommands() {
```
kuiktzecheng, line 108: Good method to have! To print out the string outputs based on what the user input is! Good idea!!

src/main/java/hana/command/ByeCommand.java
```diff
@@ -0,0 +1,44 @@
...
+     * @throws HanaException If an error occurs during command execution.
+     */
+    @Override
+    public void execute(TaskList taskList, Ui ui, Storage storage) throws HanaException {
+        ui.printBye();
+        PauseTransition pause = new PauseTransition(Duration.seconds(3));
+        pause.setOnFinished(event -> {
+            Platform.exit();
+        });
+        pause.play();
```
kuiktzecheng, line 32: Is this necessary? May I know what the rationale is for?

kuiktzecheng (commented):
Overall code looks good!!

src/main/java/hana/command/ByeCommand.java
sooyitao: It creates a delay of 3 seconds before the app terminates so that the user can read the exit message rather than exit immediately after printing the message. It's not really necessary but without the delay, there's no point in printing the message since the user wouldn't have time to read it.

PR #84, opened 2024-08-24
[t-leongchuan] iP

t-leongchuan:
# Denim 👖 
> "The name is inspired by a certain dream." – Me

Denim _sorta_ helps you but not as well as how a phone might. But hey, it:
- Spit cute responses when you enter the wrong command
- Manages your tasks such as deadlines, events and to-dos. It is too lazy to bother reminding you though. 😪 
- has probably some bugs that need to be squashed. 🐛 

To try it, all you need to so is:

1. Download the JAR file [here](https://github.com/t-leongchuan/ip/releases)
2. Run the JAR file in your terminal
3. Start using it

To start, let it spit out the possible commands and format you need to use by entering the command
`help`

Have fun using it, and use it as a refactoring exercise! I find some part of the code to be unclean 💩 , especially the Abstract `Command` class as well as its subclasses:
``` Java
public abstract class Command {
    public abstract CommandResult execute(TaskList taskList, TaskIo taskIo);
}
```    

Features:
- [x]  Manages Tasks
- [x] GUI (Coming Soon)


woke02 (commented):
LGTM. Overall, code follows coding standard, OOP is applied appropriately and code is well structured. Well Done!

src/main/java/Denim/Commands/Command.java
kimxw, line 1: Great use of packages like Command, Storage, Task etc
To follow the convention, your package names should be all lower case
eg. denim.commands

src/main/java/Denim/Commands/DeadlineCommand.java
```diff
@@ -0,0 +1,43 @@
...
+ * Represents a deadline command that can be executed.
+ */
+public class DeadlineCommand extends Command {
+    public static final String COMMAND_WORD = "deadline";
+    public static final String COMMAND_USAGE = "deadline <task> /by <date>\nWhere date is in dd/MM/yyyy HHmm";
+    private final Task deadlineTask;
+
+    public DeadlineCommand(String taskDescription, LocalDateTime deadline) {
+        deadlineTask = new Deadline(taskDescription, deadline);
+    }
```
kimxw: Add a new line between methods. Same issue at some other points.

src/main/java/Denim/Commands/HelpCommand.java
```diff
@@ -0,0 +1,31 @@
...
+import denim.storage.TaskIo;
+
+/**
+ * Represents a Help command that can be executed.
+ */
+public class HelpCommand extends Command {
+
+    public static final String COMMAND_WORD = "help";
+
+    public final String returnFormat = String.format("%s\n%s\n%s\n%s\n%s\n%s\n%s",
```
kimxw: Great work with the abstraction here. I should learn from this for my own ip.

src/main/java/Denim/Parser.java
```diff
@@ -0,0 +1,264 @@
...
+     *
+     * @param args The arguments for the delete command.
+     * @return The delete command or an invalid command if the arguments are incorrect.
+     */
+    private Command prepareDelete(String args) {
+        final Matcher matcher = MARK_UNMARK_DELETE_ARGUMENT_FORMAT.matcher(args);
+
+        if (!matcher.matches()) {
+            return new InvalidCommand("Wrong format for delete command", DeleteCommand.COMMAND_USAGE);
+        }
```
kimxw, line 246: new line after egyptian brackets to improve readibility.

src/main/java/Denim/Storage/TaskIO.java
```diff
@@ -0,0 +1,275 @@
...
+    /**
+     * Creates a save point by handling directory or file creation based on the provided status.
+     *
+     * @param status The status indicating whether the directory or file does not exist.
+     * @param sc     The scanner to receive user input for creating the necessary files.
+     * @throws DenimException If an error occurs during file or directory creation.
+     */
+    private void createSavePoint(FileStatus status, Scanner sc) throws DenimException {
+        switch (status) {
+        case DIRECTORY_DOES_NOT_EXIST:
```
kimxw: It's interesting that you chose to use an enum for file status. While it's definitely good practice, I'm curious about the point at which enums might become less advantageous compared to using else if statements.
Would like to know your opinion

src/test/java/Denim/ParserTest.java
kimxw, line 1: Correct directories setup for gradle test ✅

src/test/java/Denim/ParserTest.java
```diff
@@ -0,0 +1,94 @@
...
+import denim.commands.Command;
+import denim.commands.DeadlineCommand;
+import denim.commands.EventCommand;
+import denim.commands.InvalidCommand;
+
+public class ParserTest {
+
+    private final Parser parser = new Parser();
+    @Test
+    public void prepareEvent_validInput_returnsEventCommand() {
```
kimxw, line 20: test method names follow naming convention ✅

kimxw (commented):
Honestly great work. Going through your work has given me inspiration to improve my own work. Eventhough you said that the OOP is messy at places, I did not feel that way going through the code. Perhaps because you are more familiar with your own work you feel like things could be improved. But this project is great in my opinion!

src/main/java/Denim/Storage/TaskIO.java
t-leongchuan: Hi @kimxw , the reason why I used `enums` is just for better readability than anything else. I think it is fine to use 
`if else` statement here as there are only a few cases to consider here.

src/main/java/Denim/Commands/Command.java
t-leongchuan: Yup I'm aware it's just that still trying to learn and fiddle around with Git as the suggestion above is only local to my computer but not on GitHub.

As for why I feel that some part is messy, I think one of the better examples is in TaskIo class. Reading through the code, I realized that there were a lot of quick-fixes and separation of concern/logic was subpar for the class. 

Even though the methods are separated out to handle cases (such as file not found, or directory not found), there is still code smell in the sense that the method (handleFileNotFound() and handleDirectoryNotFound()) are still doing way too many things and is even assuming that the user-interface is in the CLI 
(Which of course, with the GUI we just used last week, that is no longer the case.)

Additionally, my initial plan was to create two classes out of TaskIo, where one is for specifically reading tasks and writing tasks. However, I figured that I was probably too lazy then to untangle the mess I made and I am currently trying to refactor the class now.

PR #100, opened 2024-08-24
[Michael Yang] iP

michaelyql:
# Terminator
> (Terminator): I will be back...

This is a chatbot inspired by the [Terminator](https://en.wikipedia.org/wiki/The_Terminator) franchise.

Features include
- Adding, deleting, marking of tasks

For example, `todo terminate aliens` creates a todo task for *terminating aliens* 🤖

In order to run the chatbot,
1. Download the latest jar release version into your preferred directory.
2. Run `java -jar Terminator` from the terminal or command line.

The entry point of the application is in the `main` function. 
```java
class Terminator {
    ...
        Terminator tChatbot = new Terminator();
        tChatbot.run();
    }
}
```
<hr />

Additional features (to be added):
- [ ] Keep track of your ingredients in your kitchen fridge
- [x] Sweep your floors for you and keep your house clean


src/main/java/terminator/Terminator.java
```diff
@@ -0,0 +1,128 @@
...
+        // Start main event loop
+        this.terminalUi.greet();
+        Scanner sc = new Scanner(System.in);
+        String input = sc.nextLine();
+        while (!input.equals("bye")) {
+            this.terminalUi.printHorizontalLine();
+            try {
+                Command command = this.parser.parse(input);
+                String response = command.execute(this.taskList.getTaskList());
+                if (command instanceof TodoCommand
```
antgyf, line 71: Would it be possible to implement this in the execute() method instead? ie. keep a static taskList field in the command class and call getTasksRemaining() on it instead. Not a big issue but something to consider.

src/main/java/terminator/command/CommandParser.java
```diff
@@ -0,0 +1,80 @@
...
+    private final Pattern todoPattern;
+    private final Pattern eventPattern;
+    private final Pattern deadlinePattern;
+    private final Pattern deletePattern;
+    private final Pattern findPattern;
+
+    /**
+     * Creates a new instance of a CommandParser.
+     */
+    public CommandParser() {
```
antgyf, line 39: Might be more efficient to keep these patterns as a static field and initialise on declaration.

src/main/java/terminator/command/CommandParser.java
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * Parses the user input and returns an instance of a concrete implementation of the Command class,
+     * which can be executed to perform the desired operation.
+     *
+     * @param input The user input to parse
+     * @return An instance of the Command class.
+     */
+    public Command parse(String input) {
+        Matcher matcher;
+        if (input.isEmpty()) {
```
antgyf, line 51: for many layers of if else statements it may be faster to use a switch statement!

src/main/java/terminator/Terminator.java
```diff
@@ -0,0 +1,128 @@
...
+     *
+     * If the data was read successfully, initiate the main event loop, which can
+     * handle commands from the user through the command line. <br/><br/>
+     *
+     * The application exits when the user enters {@code bye} into the command line.
+     */
+    private void runTerminalUi() {
+        // Load data from storage
+        try {
+            this.storage.loadDataFromFile(this.taskList);
```
antgyf: nit: can remove unnecessary use of this keyword. Should only use this keyword if there is ambiguity in code.

antgyf (commented):
Overall LGTM. Just a few small nits here and there!

src/main/java/terminator/Terminator.java
michaelyql: I feel that keeping a static `taskList` field in the `Command` class could lead to poorer separation of concerns, as the `Command` class is only intended to represent a command by the user, and hence it should not be concerned with how data is being stored.

src/main/java/terminator/command/CommandParser.java
michaelyql: I decided to stick with `if/else` statements because `switch` statements require a target variable that is either a byte, short, char, int (and their wrappers), or String or enum. Switch statements can't match objects, so in this case it wouldn't make sense to try to match the `Matcher` object.

I feel that regex patterns are more robust and less error-prone in general for parsing user inputs, but I agree that it has performance downsides and is slower than regular switch statements.

src/main/java/terminator/command/DeadlineCommand.java
```diff
@@ -0,0 +1,51 @@
...
+        super(input);
+    }
+
+    /**
+     * Creates a new DeadlineTask with the given input and adds it to the task list.
+     *
+     * @param todoList The task list.
+     * @throws TerminatorException if the description of the deadline task is empty.
+     */
+    @Override
```
xGladiate, line 29: When writing Javadocs for overridden methods, the `@inheritDoc` tag can be used to reuse the header comment from the parent method but with further modifications e.g., when the method has a slightly different behavior from the parent method. _Else, if f I am not wrong:_ The JavaDoc tool will use the comment from the overriding method.

src/main/java/terminator/task/Storage.java
```diff
@@ -0,0 +1,246 @@
...
+                    break;
+                case EVENT:
+                    parseEvent(desc);
+                    break;
+                case DEADLINE:
+                    parseDeadline(desc);
+                    break;
+                default:
+                    // We shouldn't reach this condition since we already checked the TaskType beforehand
+                    break;
```
xGladiate: Good job on not making any common mistakes! 💯

src/main/java/terminator/Main.java
```diff
@@ -0,0 +1,110 @@
...
+
+    private Image userImage = new Image(this.getClass().getResourceAsStream("/images/user.jpg"));
+    private Image dukeImage = new Image(this.getClass().getResourceAsStream("/images/terminator.jpg"));
+
+    /**
+     * Starts the GUI for the chatbot.
+     *
+     * @param stage The JavaFX Stage.
+     */
+    @Override
```
xGladiate, line 27: Good job grouping the code, which helps to provide clear segmentation of the different segments of code. 
**_Improvements_**: Might want to provide clear and concise explanations of what each section of code is doing. 

For example: 
``` java
    // Configure user input and send button
    userInput.setPrefWidth(325.0);
    sendButton.setPrefWidth(55.0);
```

src/main/java/terminator/Terminator.java
```diff
@@ -0,0 +1,128 @@
...
+
+import terminator.command.Command;
+import terminator.command.CommandParser;
+import terminator.command.DeadlineCommand;
+import terminator.command.DeleteCommand;
+import terminator.command.EventCommand;
+import terminator.command.TerminatorException;
+import terminator.command.TodoCommand;
+import terminator.task.Storage;
+import terminator.task.TaskList;
```
xGladiate: Good job for grouping the imports in a clear manner!

xGladiate (commented):
Overall, there is no visible violation of coding standards, keep up the good work! 💯

PR #112, opened 2024-08-25
[Qiu Jinhang] iP

jhqiu21:
# EchoBot
> "The key is not to prioritize what's on your schedule, but to schedule your priorities." -Stephen Covey ([source](https://www.brainyquote.com/authors/stephen-covey-quotes))

EchoBot is your ultimate task manager, designed to free up your mental space so you can focus on what truly matters. With EchoBot you can effortlessly keep track of your tasks without the hassle of remembering every detail. :star_struck: It’s:

* Easy to learn so you can get started right away.
* Blazing fast for quick and efficient task management.

Getting started is as simple as:

1. Downloading it from [here](https://github.com/jhqiu21/ip/releases/download/A-Jar/echo.jar).
1. Open a terminal.
1. Navigate to the folder with `echo.jar` file using the cd command.
1. Run the command 
    ```
    java -jar echo.jar.
    ```
1. Adding your tasks.
1. Letting EchoBot take care of the rest. :wink:

**Features:**

- [x] Task management
- [x] Deadline tracking
- [ ] Customizable themes (coming soon)
- [ ] Reminder notifications (coming soon)

```java
public class EchoBot {
        EchoBot bot = new EchoBot();
        bot.run();
    }
}
```


src/test/java/echo/DeadlineTest.java
kimxw, line 1: Not sure but I was wondering if the test directory should be under root as opposed to src

src/main/java/echo/Storage.java
kimxw, line 1: You could consider using more packages to make the codebase more maintainable. For example, Task and related classess could be put in a package called tasks. You could have a package called storage for classess that read and write from the memory/ data text file.

src/main/java/echo/Parser.java
```diff
@@ -0,0 +1,111 @@
...
+        // parse the command
+        String[] cmdParts = userInput.split(" ", 2);
+        String command = cmdParts[0].toUpperCase();
+
+        // set isExit flag to indicate bot exit
+        // boolean isExit = false;
+
+        try {
+            switch (Command.valueOf(command)) {
+            case BYE:
```
kimxw: Correct indentation for switch statement (as per this course :) because I have seen many people miss it (including myself) oof

src/main/java/echo/ToDo.java
```diff
@@ -0,0 +1,29 @@
...
+ * This class extends the Task class and implements Serializable for object serialization.
+ */
+public class ToDo extends Task implements Serializable {
+
+    /**
+     * Creates a task to be completed.
+     *
+     * @param task without any date/time attached to it.
+     */
+    public ToDo(String task) {
```
kimxw: Perhaps use a 'description' as a variable name instead of task as it gives a slightly false impression that task is of type Task. Same comment for the parent Task class' attributes.

src/main/java/echo/ToDo.java
```diff
@@ -0,0 +1,29 @@
+package echo;
+
+import java.io.Serializable;
+
+/**
+ * Represents a task to be completed.
+ * This class extends the Task class and implements Serializable for object serialization.
+ */
+public class ToDo extends Task implements Serializable {
```
kimxw: Nice use of this interface. But i guess we dont have to explicitely say that it implements Serializable here because the parent Task class aready does that (?)

src/main/java/DialogBox.java
haleemairfan, line 1: Hi! You could shift your files handling the GUI into a separate directory for better organisation!

src/main/java/echo/Command.java
haleemairfan, line 1: great use of enums i personally did not think of that :)

src/main/java/echo/TaskList.java
```diff
@@ -0,0 +1,118 @@
...
+        for (int i = 0; i < this.allTasks.size(); i++) {
+            tasks += (i + 1) + ". " + this.allTasks.get(i).toString();
+            tasks += (i == this.allTasks.size() - 1) ? "" : "\n";
+        }
+        return "There are " + this.allTasks.size() + " tasks in the list:"
+                + "\n" + tasks;
+    }
+
+    /**
+     * Marks the target task in the list based on the index.
```
kimxw: JavaDoc convention of starting the header comment with "Marks" "Returns" etc ✅

src/main/java/echo/Deadline.java
```diff
@@ -0,0 +1,42 @@
...
+public class Deadline extends Task implements Serializable {
+    private LocalDateTime deadline;
+
+    /**
+     * Creates task that need to be done before a specific date/time.
+     *
+     * @param task task information of the deadline task.
+     * @param deadline the time by which the task needs to be completed,
+     *                 in the format "yyyy/MM/dd HH:mm".
+     * @throws DateTimeParseException if the deadline string is not in the correct format.
```
haleemairfan: Your javadoc comments are very detailed!

src/main/java/echo/Parser.java
```diff
@@ -0,0 +1,111 @@
...
+     */
+    public static String parse(String userInput, TaskList allTasks) throws DukeException {
+        // parse the command
+        String[] cmdParts = userInput.split(" ", 2);
+        String command = cmdParts[0].toUpperCase();
+
+        // set isExit flag to indicate bot exit
+        // boolean isExit = false;
+
+        try {
```
haleemairfan: neat use of enums!

src/main/java/echo/Parser.java
```diff
@@ -0,0 +1,111 @@
+package echo;
+
+
+/**
+ * The Parser class is responsible for interpreting user commands
+ * and performing the corresponding actions on the task list.
+ */
+public class Parser {
+
+    public static String listAllTask(TaskList allTasks) {
```
kimxw: Method name 'listAllTasks' might be better + this would make it more consistent with the 'allTasks' variable as well

src/main/java/echo/TaskList.java
```diff
@@ -0,0 +1,118 @@
...
+
+import java.io.Serializable;
+import java.util.ArrayList;
+
+/**
+ * Represents a list of Task objects.
+ * It provides methods to operate on a target Task object in the list.
+ */
+public class TaskList implements Serializable {
+    private ArrayList<Task> allTasks = new ArrayList<>();
```
kimxw: I generally checked all classes and seems like you didnt make any of the attributes public. nice! ✅

src/test/java/echo/DeadlineTest.java
```diff
@@ -0,0 +1,31 @@
+package echo;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.junit.jupiter.api.Test;
+
+public class DeadlineTest {
+    @Test
+    public void test_creat_deadline() {
```
kimxw: test methods dont follow the naming convention

Underscores may be used in test method names using the following three part format `featureUnderTest_testScenario_expectedBehavior()`

e.g. `sortList_emptyList_exceptionThrown()`, `getMember_memberNotFound_nullReturned`

src/test/java/echo/DeadlineTest.java
```diff
@@ -0,0 +1,31 @@
...
+    @Test
+    public void test_mark_deadline() {
+        try {
+            Task deadlineA = new Deadline("dA", "2023/03/23 23:59");
+            Task deadlineB = Task.createTask("deadline dA /by 2023/03/23 23:59");
+            deadlineA.setMark();
+            deadlineB.setMark();
+            assertEquals(deadlineA.toString(), deadlineB.toString());
+        } catch (DukeException e) {
+            System.out.println(e.getMessage());
```
kimxw: I think in general it is not adviced to add print statements to tests. I did a quick google search and maybe the JUnit `assertThrows` will be a better way to detect exception and pass/fail the test accoridngly.

kimxw (commented):
Some naming standard violations and things to take note of when testing with exceptions. Other than that, LGTM!

PR #113, opened 2024-08-25
[MinkhantEain] ip

MinkhantEain:
# Luke
> "I am your father" - [Darth Vader](https://en.wikipedia.org/wiki/Darth_Vader)

The name Luke was chosen for the following reasons:
+ Similar to Duke
- ~~Easy~~ Simple change from Duke
* Same first name as __Luke Skywalker__ :baby: 

What can it do for you?
1. It can track your list of things to do
2. It can show the deadline :nerd_face: 
3. It can show the task timeline :open_mouth: 

All this functionality at **SGD__0.00__!**

Features:
- [x] Basic task tracking functionality
- [ ] GUI implementation (Coming Soon)
- [ ] Push notification (Coming Soon)

Just for info, the `main` method looks like this:
```java
public static void main(String[] args) throws IOException {
        new Luke("./data/Luke.txt").run();
    }
```


src/main/java/presentation/Ui.java
```diff
@@ -0,0 +1,87 @@
...
+                        + "| |__ | |_| |   <  __/\n"
+                        + "|____| \\__,_|_|\\_\\___|\n";
+        System.out.println("Hello from\n" + logo);
+        System.out.println("Yo! I'm Luke.\nHow's it hanging?");
+    }
+
+    /**
+     * Shows closing dialog when the application ends.
+     */
+    public void closingDialog() {
```
itsnotsherm: Would it be clearer to have your method be named "closeDialog" without the adverb?

src/main/java/tasks/Task.java
```diff
@@ -0,0 +1,70 @@
+package tasks;
+
+import java.util.Objects;
+
+public abstract class Task {
+    private boolean completed;
```
itsnotsherm: I like your naming. Should this boolean be named isComplete instead?

src/main/java/tasks/Event.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return String.format("[E][%s] %s (from: %s to: %s)", super.completedStringRepresentation(), super.getDescription(), from, to);
+    }
+
+    public static Event load(String input) {
+        String[] parameters = input.split("\\|");
+        String[] dateRange = parameters[3].split("to");
```
itsnotsherm, line 77: Perhaps a better naming to include a plural noun instead so that it is clearly a collection of Strings?

src/main/java/tasks/Task.java
```diff
@@ -0,0 +1,70 @@
...
+
+    public void mark() {
+        this.completed = true;
+    }
+
+    public void unMark() {
+        this.completed = false;
+    }
+
+    public String completedStringRepresentation() {
```
itsnotsherm: Should this be renamed to a verb that represents the method?

src/main/java/utility/TaskList.java
```diff
@@ -0,0 +1,99 @@
...
+
+import java.io.IOException;
+import java.util.ArrayList;
+
+/** Wrapper class for tasks recorded by Luke. */
+public class TaskList {
+    private final ArrayList<Task> tasks;
+
+    /** Returns the size of tasks. */
+    public int getTasksSize() {
```
itsnotsherm: Would it be clearer if this is named getTaskListSize instead?

src/main/java/tasks/Task.java
```diff
@@ -0,0 +1,70 @@
...
+        if (!completed) {
+            return " ";
+        } else {
+            return "X";
+        }
+    }
+
+    public int intComplete() {
+        return completed ? 1 : 0;
```
itsnotsherm: Any idea why you do this instead of "return completed == true"?

itsnotsherm (commented):
LGTM, just a few nits to fix with regards to variable and method names

PR #130, opened 2024-08-25
[tzuennn] iP

tzuennn:
# Nugget
> "Your mind is for having ideas, not holding them." - David Allen [source](https://dansilvestre.com/productivity-quotes/)

Nugget frees your mind of having to remember things you need to do. It's,
- text-based
- easy to learn
- ~~FAST~~ *SUPER* FAST to use

All you need to do is,
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you 😄 

And it is **FREE!**

Features:
- [x] Managing tasks
- [ ] Managing deadlines (coming soon!)
- [ ] Reminders (coming soon!)

```java
public class Main {
    }
}
```


src/main/java/nugget/Parser.java
```diff
@@ -0,0 +1,122 @@
...
+            int index = Integer.parseInt(details.trim()) - 1;
+            if ((index < 0) || (index >= tasks.size())) {
+                throw new InvalidTaskNumberException();
+            }
+            return index;
+        } catch (NumberFormatException e) {
+            throw new EmptyDescriptionException();
+        }
+    }
+
```
jinhanjinhan: I like how you have a method specifically for parsing index to check whether the index is valid!

src/main/java/nugget/Storage.java
```diff
@@ -0,0 +1,81 @@
+package nugget;
```
jinhanjinhan, line 1: Not sure where to put this,  but I am thinking would it be better to separate the classes to separate sub packages? Maybe something like nugget.exception and nugget.command?

src/main/java/nugget/Storage.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * Loads tasks from the file specified by the file path.
+     * Each line in the file represents a task, which is converted back to a {@code Task} object.
+     *
+     * @return An {@code ArrayList} of tasks loaded from the file.
+     */
+    public ArrayList<Task> loadTasks() {
+        ArrayList<Task> tasks = new ArrayList<>();
```
jinhanjinhan, line 36: Great naming of "tasks" instead of "task"

src/main/java/nugget/TaskList.java
```diff
@@ -0,0 +1,53 @@
...
+    public TaskList() {
+        this.tasks = new ArrayList<>();
+    }
+
+    public TaskList(ArrayList<Task> tasks) {
+        this.tasks = tasks;
+    }
+
+    public void addTask(Task task) {
+        tasks.add(task);
```
jinhanjinhan, line 22: Once again, great differentiation between "task" and "tasks"!

src/main/java/nugget/Ui.java
```diff
@@ -0,0 +1,69 @@
+package nugget;
+
+import nugget.Task;
+
+import java.util.ArrayList;
+
+public class Ui {
+
+    public void line() {
```
jinhanjinhan: For this part, I think it would be good if you use printLine() instead of print() to make it in the form of an action

src/test/java/nugget/TaskListTest.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @Test
+    void markTask_shouldMarkTaskAsCompleted() {
+        taskList.addTask(todoTask);
+        taskList.markTask(0);
+        assertTrue(taskList.getTask(0).isCompleted(), "Task should be marked as completed");
+    }
+
+    @Test
```
jinhanjinhan, line 49: For the test methods, maybe a better name would be in the form of unitBeingTested_descriptionOfTestInputs_expectedOutcome (as specified from the notes)?

jinhanjinhan (commented):
Overall the naming is great! Most parts follow the naming convention, except for the testing methods.

src/main/java/nugget/MarkTaskCommand.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    /**
+     * Executes the mark task command, marking the specified task as completed.
+     *
+     * @param tasks The list of tasks where the task will be marked as completed.
+     * @param ui The user interface that displays messages to the user.
+     * @param storage The storage that handles the saving of tasks.
+     * @throws NuggetException If the provided task index is invalid.
+     */
```
Wnayar: Nice descriptive header comments for public methods.

src/main/java/nugget/Ui.java
```diff
@@ -0,0 +1,69 @@
...
+        System.out.println("________________________________________");
+    }
+    public void printIntro() {
+        System.out.println("________________________________________");
+        System.out.println("Hello! I'm nugget.Nugget");
+        System.out.println("What can I do for you?");
+        System.out.println("________________________________________");
```
Wnayar: Awesome looking UI. Have you considered using a variable for the line breaks?

src/main/java/nugget/FindTaskCommand.java
```diff
@@ -0,0 +1,20 @@
...
+
+public class FindTaskCommand implements Command {
+    private String searchQuery;
+
+    public FindTaskCommand(String searchQuery) {
+        this.searchQuery = searchQuery;
+    }
+
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws NuggetException {
```
Wnayar: I think the abstraction of each command is great. Have you considered giving a more descriptive function name, because in the context of FindTaskCommand now its okay because its the only function. But say in the future if you add to this class another function then it may not be as clear what the function purpose is. But overall awesome work!

src/main/java/nugget/TaskList.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public void markTask(int index) {
+        tasks.get(index).markTask();
+    }
+
+    public void unmarkTask(int index) {
+        tasks.get(index).unmarkTask();
+    }
+
+    public int size() {
```
Wnayar, line 37: Great naming conventions. I think here size could be replace with something more verb like. Perhaps getSize().

src/test/java/nugget/TaskListTest.java
Wnayar: I agree with this comment.

Wnayar (commented):
LGTM. Just some nits related to CS2103T coding standards to fix.

src/test/java/nugget/TaskListTest.java
tzuennn: Thank you for pointing that out!

PR #132, opened 2024-08-25
[Jiale] iP

jialegitgud:
# Peter
> "Coding like poetry should be short and concise."  - Santosh Kalwar

Peter is 
- text-based
- user friendly
- FAST AND SUPER EASY TO USE!

To start using Peter, you should,
1. download it from [here](https://github.com/jialegitgud/ip/tree/master).
2. double-click the application icon.
3. add your tasks to the application.
4. and let it manage your tasks for you! 💯 

And it is completely **FREE TO USE**!

_Features:_

- [X]  Managing tasks
- [X]  Managing deadlines
- [X]  Managing events
- [ ] Providing reminders

_______________________________________________________________________________________________________________________________________

```
    new Peter().run();
}
```


src/main/java/Exceptions/UnknownCommandException.java
```diff
@@ -0,0 +1,10 @@
+package exceptions;
+
+/**
+ * Raised exception on unknown command input.
```
MAOXIONGKAI: Just a suggestion 🤔 Would it be better to change the javadoc description from "Raised" to "Raises" as according to the [javadoc guideline in Week 3 topics](https://nus-cs2103-ay2425s1.github.io/website/schedule/week3/topics.html)

src/main/java/Peter/Peter.java
```diff
@@ -0,0 +1,41 @@
...
+
+import utilities.CommandParser;
+import utilities.Storage;
+import utilities.TaskList;
+import utilities.UI;
+
+public class Peter {
+    public static final String FILE_PATH = "tasks/data.txt";
+    private Storage store;
+    private TaskList tl;
```
MAOXIONGKAI: Avoid abbreviation for variables and spell them out in full, "tasks" or "tasklist" would be a better name over here

src/main/java/Tasks/Deadlines.java
```diff
@@ -0,0 +1,29 @@
+package tasks;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+
+public class Deadlines extends Task {
+    private LocalDateTime deadlineInfo;
+    private static final DateTimeFormatter PATTERN_DISPLAY = DateTimeFormatter.ofPattern("MMM dd yyyy HH:mm");
+    private static final DateTimeFormatter PATTERN_READ = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm");
```
MAOXIONGKAI, line 12: I like the use of SCREAMING_SNAKE_CASE for constant over here, which makes the code make a lot more readable

src/main/java/Utilities/Storage.java
```diff
@@ -0,0 +1,135 @@
...
+                    tasks.add(new ToDos(splits[2]));
+                    break;
+                case "D":
+                    tasks.add(new Deadlines(splits[2], splits[3].strip()));
+                    break;
+                case "E":
+                    tasks.add(new Event(splits[2], splits[3].strip(), splits[4].strip()));
+                    break;
+                default:
+                    break;
```
MAOXIONGKAI, line 53: I like how you include a default case here for clarity and code safety purpose! 😄

src/main/java/Utilities/Storage.java
```diff
@@ -0,0 +1,135 @@
...
+        ArrayList<String> fileContents = new ArrayList<>();
+
+        try {
+            Scanner fileScanner = new Scanner(this.file);
+            while (fileScanner.hasNextLine()) {
+                fileContents.add(fileScanner.nextLine());
+            }
+            fileScanner.close();
+
+            String s = fileContents.get(index);
```
MAOXIONGKAI, line 93: Would it better to replace "s" with something more specific? I don't know what exactly is "s" just by looking at this piece of code, maybe the future you won't know about this anymore either, so a name that can explains the variable itself is probably better?

MAOXIONGKAI (commented):
LGTM, just a few nits to fix. I like the implementation of your code, good luck on your project! 😄

src/main/java/Peter/DialogBox.java
```diff
@@ -0,0 +1,60 @@
...
+
+        dialog.setText(text);
+        displayPicture.setImage(img);
+    }
+
+    /**
+     * Flips the dialog box such that the ImageView is on the left and text on the right.
+     */
+    private void flip() {
+        ObservableList<Node> tmp = FXCollections.observableArrayList(this.getChildren());
```
sayomaki, line 45: The variable naming is a bit ambiguous here, but good that you have added a comment to the function so that people know what it does!

src/main/java/Peter/MainWindow.java
```diff
@@ -0,0 +1,54 @@
+package peter;
+
+import javafx.fxml.FXML;
+import javafx.scene.control.Button;
+import javafx.scene.control.ScrollPane;
+import javafx.scene.control.TextField;
+import javafx.scene.image.Image;
+import javafx.scene.layout.AnchorPane;
+import javafx.scene.layout.VBox;
+/**
```
sayomaki, line 11: Maybe it would be nice to have a space between the imports and the javadoc?

src/main/java/Peter/Peter.java
```diff
@@ -0,0 +1,23 @@
...
+import utilities.Storage;
+import utilities.TaskList;
+
+public class Peter {
+    public static final String FILE_PATH = "tasks/data.txt";
+    private Storage store;
+    private TaskList tasklist;
+
+    public Peter() {
+        this.store = new Storage(FILE_PATH);
```
sayomaki, line 16: Since `store` is not ambiguous here, omitting the `this` would be okay (refer to code guidelines under "Avoid unnecessary use of this with fields.")

src/main/java/Tasks/Deadlines.java
```diff
@@ -0,0 +1,29 @@
...
+
+    public Deadlines(String name, String details) {
+        super(name);
+        this.deadlineInfo = LocalDateTime.parse(details, PATTERN_READ);
```
sayomaki, line 16: Here the reference to the field `deadlineInfo` is not ambiguous, can possibly omit `this` keyword

src/main/java/Tasks/Task.java
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * Toggles isDone as true.
+     */
+    public void markAsDone() {
+        this.isDone = true;
+    }
+
+    /**
+     * Toggles isDone as true.
```
sayomaki: I don't think the javadoc matches what the method actually does here :D

src/main/java/Tasks/Task.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    /**
+     * Returns string representation of the isDone status.
+     */
+    public String getTaskStatus() {
+        return (isDone ? "X" : " ");
+    }
+
+    public abstract String getWriteTaskInfo();
```
sayomaki: Since this is an abstract method (which needs to be implemented by subclasses), would it be better to include some javadoc on what should the output look like? (or maybe what the method should do?)

src/main/java/Utilities/CommandParser.java
```diff
@@ -0,0 +1,96 @@
...
+import tasks.ToDos;
+
+public class CommandParser {
+    /**
+     * Parses and performs actions for the given command.
+     *
+     * @param input String command for processing.
+     * @param tl TaskList class for maintaining tasks.
+     * @param store Storage class for writing to files.
+     * @return String value for user response.
```
sayomaki, line 47: Good use of javadoc annotations to describe what each parameter is and the return value of the method!

src/main/java/Utilities/Storage.java
```diff
@@ -0,0 +1,135 @@
...
+                } else {
+                    tasks.get(tasks.size() - 1).markAsDone();
+                }
+            }
+            fileScanner.close();
+        } catch (FileNotFoundException e) {
+            // Admin Message for Generation
+            System.out.println("File has been generated.");
+            try {
+                this.file.createNewFile();
```
sayomaki, line 65: `this` here can be omitted as the reference is clear.

src/main/java/Utilities/UI.java
```diff
@@ -0,0 +1,52 @@
+package utilities;
+
+import tasks.Task;
+
+// Handles all system interactions
+public class Ui {
```
sayomaki: Maybe you can add some javadoc for the public methods of this class? :)

src/test/java/utilities/TaskListTest.java
```diff
@@ -0,0 +1,33 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import org.junit.jupiter.api.Test;
+
+// Assumes the data file is based on the default generated by runtest
+public class TaskListTest {
+    private static final String FILE_PATH = "tasks/data.txt";
+    private static final String ACTUAL_STRING = "1.[T][ ]  borrow book \n"
+            + "2.[E][X]  project meeting from: Dec 29 2024 12:00 to: Dec 29 2024 17:00 \n"
+            + "3.[T][ ]  borrow notes";
+    private Storage store = new Storage(FILE_PATH);
```
sayomaki, line 17: I think there can be a space above this line as this is separate from the 2 constants above.

sayomaki (commented):
Overall coding standard mostly complies with the coding guidelines, but there are some minor issues (see comments below!)

PR #137, opened 2024-08-25
[Dylan Wong] iP

Potatoad88:
### **Chatbot**

> "If you fail to plan, you are planning to fail!" - Benjamin Franklin ([source](https://www.goodreads.com/quotes/460142-if-you-fail-to-plan-you-are-planning-to-fail))

_Features:_

- [x] Add/Delete/List/Save tasks
- [ ] Reminders

_Types of tasks:_

- Todo
- Deadline
- Event

_Steps to use:_

1. Submit/Type task details
2. Get confirmation message it has been added
3. Fetch and manage your tasks whenever😀

Below is my `main` method:
```java
    new DemureBot().run();
}

### CODE REVIEW

File: src/main/java/demurebot/command/DeleteCommand.java [hunk 1]
```diff
@@ -0,0 +1,46 @@
...
+
+import demurebot.task.Task;
+import demurebot.task.TaskList;
+import demurebot.ui.Ui;
+
+/**
+ * Represents a command to delete a task from the task list.
+ */
+public class DeleteCommand extends Command {
+    private final String remainder;
```
rayray39: Perhaps a more intuitive variable name to represent the index of the task to be deleted?

src/main/java/demurebot/command/MarkCommand.java
```diff
@@ -0,0 +1,46 @@
...
+        try {
+            int index = Integer.parseInt(remainder) - 1;
+            Task task = list.getTask(index);
+            task.markAsDone();
+            ui.displayMarkTask(task);
+        } catch (NumberFormatException e) {
+            System.out.println("Please enter a positive integer after mark.\n");
+        } catch (IndexOutOfBoundsException e) {
+            System.out.println("Invalid index: " +
+                (Integer.parseInt(remainder)) +
```
rayray39: Is there a need to parse remainder into an integer, since remainder is already a string?

src/main/java/demurebot/command/MarkCommand.java
```diff
@@ -0,0 +1,46 @@
...
+    public void execute(TaskList list, Ui ui) {
+        try {
+            int index = Integer.parseInt(remainder) - 1;
+            Task task = list.getTask(index);
+            task.markAsDone();
+            ui.displayMarkTask(task);
+        } catch (NumberFormatException e) {
+            System.out.println("Please enter a positive integer after mark.\n");
+        } catch (IndexOutOfBoundsException e) {
+            System.out.println("Invalid index: " +
```
rayray39: Should the line breaks happen before the '+' operator, as per the Java coding standard?

src/main/java/demurebot/storage/Storage.java
```diff
@@ -0,0 +1,190 @@
...
+    /**
+     * Returns a demure.Task created from saved data.
+     *
+     * @param line Line of text read from saved data.
+     * @return demure.Task created from saved data.
+     * @throws DemureBotException If the saved task has invalid format.
+     */
+    private Task getTask(String line) throws DemureBotException {
+        String[] split = line.split(" \\| ");
+        validateTaskFormat(split, 3);
```
rayray39: Should the ```validateTaskFormat(split, 3)``` be inside the ```case "T"```, similar to how deadline and event are validated first before creation?

rayray39 (commented):
I like how the different commands are separated into their own classes, makes it very clean and easy to understand 👍🏻 
Overall, the code was easy to understand but just some minor issues about coding standard violations.

src/main/java/demurebot/DemureBot.java
```diff
@@ -0,0 +1,96 @@
...
+        // check if data folder exists if not create it
+        String folderPath = "./data";
+        this.storage.checkFolder(folderPath);
+
+        // check if saved data exists if not create it
+        String filePath = "./data/tasks.txt";
+        boolean savedDataExists = this.storage.checkFile(filePath);
+
+        // load saved data if exists
+        if (savedDataExists) {
```
stephaneleong: Perhaps a better naming convention for this variable is to start with 'is' or 'has' to indicate that it is a boolean variable. a suggestion "hasSavedData"

src/main/java/demurebot/DemureBot.java
```diff
@@ -0,0 +1,96 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Scanner;
+
+/**
+ * Main class for the DemureBot application.
+ * This class handles the initialization and execution of the chatbot.
+ */
+public class DemureBot {
+    private TaskList list;
```
stephaneleong, line 18: Perhaps using a noun in the variable would be clearer to users. makes it less ambiguous to readers

src/main/java/demurebot/command/DeleteCommand.java
stephaneleong: a suggestion would be to think of a variable name that would be understandable within the scope of the class

src/main/java/demurebot/command/FindCommand.java
```diff
@@ -0,0 +1,36 @@
+package demurebot.command;
+
+import demurebot.task.TaskList;
+import demurebot.ui.Ui;
+
+/**
+ * Represents a command to find tasks with a keyword.
+ */
+public class FindCommand extends Command {
+    private final String keyword;
```
stephaneleong, line 10: Nice Job on giving a keeping the variable name simple 👍🏽

stephaneleong (commented):
Overal good job on keeping to good naming convention 💯 . Just some nix to fix

PR #138, opened 2024-08-25
[LimErnest] iP

LimErnest:
# Echo Chat Bot

The IDE I'm using is [Intellij](https://www.jetbrains.com/idea/)

> Quoting something

My **FREE** and **AMAZING** chat bot has the following features:
* add 3 types of task: to do, deadline and event
* use `list` to list all of the tasks added
* mark or unmark task that are completed
* ~~refuse to compile~~ 💀 

> [!WARNING]
> Use at your own risk. 

All you need to do is,
1. download
2. run the program
3. add tasks
4. let it manage your tasks for you  👍 

Increments:
- [x] Checkstyle
- [ ] Level-10 (coming soon)
- [ ] A-Varargs (coming soon)

Here is a sneak peak of my code:

```java

public void bye() {
        System.out.println(LINE_BREAK);
        System.out.println("Goodnight! Hope to see you again soon!");
        System.out.println("      |\\      _,,,---,,_\n" +
                           "ZZZzz /,`.-'`'    -.  ;-;;,_\n" +
                           "     |,4-  ) )-,_. ,\\ (  `'-'\n" +
                           "    '---''(_/--'  `-'\\_)");
        System.out.println(LINE_BREAK);
    }

```


src/main/java/echo/exception/EchoException.java
```diff
@@ -0,0 +1,18 @@
...
+
+    /**
+     * Constructor for EchoException object
+     *
+     * @param errorMessage message to be displayed when the error is caught
+     */
+    public EchoException(String errorMessage) {
+        super(errorMessage);
+    }
+}
```
chongsun2002, line 18: Can this exception be more specific, and have other exceptions that inherit from it? This may help in adding new exceptions that have different behaviors.

chongsun2002 (commented):
Your code looks very good! All the components are well abstracted. I don't have much to nit on style, so I only have some suggested improvements. The code fulfils its purposes. LGTM!

src/main/java/echo/main/Echo.java
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    public String greetUser() {
+        try {
+            this.storage.loadStorage(this.parser, this.taskList);
+            return ui.greet();
+        } catch (IOException e) {
+            return ui.printErrorMessage(e);
+        }
+    }
```
C5hives, line 49: Is this method doing more than what its name suggests? You can consider breaking this method into two separate methods.

The method `greetUser()` currently has a side effect of loading file contents into your `taskList` and needs to handle a possible `IOException`—which is thrown from the `this.storage,loadStorage()` method and not the `ui.greet()` method.

This behavior will not be expected when other classes call `greetUser()` as the current method name makes no mention of loading tasks from a file.

You can also consider keeping the current logic as is, but renaming this method to something like `initialise()` to reduce the chances of misunderstandings when reading the code.

src/main/java/echo/main/Echo.java
```diff
@@ -0,0 +1,233 @@
...
+    /**
+     * Constructor to create Echo Object
+     *
+     * @param filePath String that contains the file path to store or load the list of task
+     */
+    public Echo(String filePath) {
+        this.ui = new Ui();
+        this.taskList = new TaskList();
+        this.parser = new Parser();
+        this.storage = new Storage(filePath);
```
C5hives, line 39: Optional, but you can consider adhering to the advanced [coding standards](https://se-education.org/guides/conventions/java/index.html#variables) set out by this course, which is
- Avoid unnecessary use of this with fields. Use the `this` keyword only when a field is shadowed by a method or constructor parameter.

Since `ui`, `taskList`, `parser` and `storage` are not shadowed by any other parameters here, you can omit the `this` keyword (this would be applicable for every other class as well).

src/main/java/echo/task/ToDos.java
```diff
@@ -0,0 +1,40 @@
+package echo.task;
+
+/**
+ * Represents a to-do task with descriptions
+ *
+ * @author Ernest Lim
+ */
+public class ToDos extends Task {
```
C5hives, line 8: You can consider renaming the `ToDos` class (and every other instance of it) to be singular as it represents only 1 task. It would then adhere to the [coding standard](https://se-education.org/guides/conventions/java/index.html#naming), which states that:
- Plural form should be used on names representing a collection of objects.

src/main/java/echo/main/Echo.java
```diff
@@ -0,0 +1,233 @@
...
+                return handleEventCommand(description);
+            case "delete":
+                return handleDeleteCommand(description);
+            case "find":
+                return handleFindCommand(description);
+            case "bye":
+                return handleByeCommand();
+            default:
+                throw new EchoException("Sorry! I don't get what you mean. Try again with the list of commands above.");
+            }
```
C5hives, line 195: Nice work on using separate methods to extract out the logic used for each command. This section of code is very easy to read! 🤩

src/main/java/echo/tasklist/TaskList.java
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * Represents a list of task
+ *
+ * @author Ernest Lim
+ */
+public class TaskList {
+
+    /** ArrayList of the tasks added to TaskList by the user */
+    private ArrayList<Task> taskList;
```
C5hives, line 17: Perhaps a more intuitive variable name here? You can consider renaming `taskList` to `tasks` to indicate that it is a collection of `Task` objects.

src/main/java/echo/tasklist/TaskList.java
```diff
@@ -0,0 +1,124 @@
...
+     * @param keyword word to search for in the list of task
+     * @return an arrayList of all the task with description containing the keyword
+     */
+    public ArrayList<Task> findTask(String keyword) {
+        ArrayList<Task> resultArray = new ArrayList<>();
+        for (Task task : this.taskList) {
+            String description = task.getDescription();
+            if (description.contains(keyword)) {
+                resultArray.add(task);
+            }
```
C5hives, line 120: Could this logic be extracted out into the `Task` class directly? This could help with the abstraction of the `Task` class and hide its internal implementation details.

C5hives (changes requested):
The use of OOP has made the project easy to follow and read. Nice job!

src/main/java/echo/main/Echo.java
C5hives: Optional, but you can consider adhering to the advanced [coding standards](https://se-education.org/guides/conventions/java/index.html#variables) set out by this course, which is
- Avoid unnecessary use of this with fields. Use the `this` keyword only when a field is shadowed by a method or constructor parameter.

Since `ui`, `taskList`, `parser` and `storage` are not shadowed by any other parameters here, you can omit the `this` keyword (this would be applicable for every other class as well).

src/main/java/echo/main/Echo.java
C5hives: Is this method doing more than what its name suggests? You can consider breaking this method into two separate methods.

The method `greetUser()` currently has a side effect of loading file contents into your `taskList` and needs to handle a possible `IOException`—which is thrown from the `this.storage,loadStorage()` method and not the `ui.greet()` method.

This behavior will not be expected when other classes call `greetUser()` as the current method name makes no mention of loading tasks from a file.

You can also consider keeping the current logic as is, but renaming this method to something like `initialise()` to reduce the chances of misunderstandings when reading the code.

src/main/java/echo/main/Echo.java
C5hives: Nice work on using separate methods to extract out the logic used for each command. This section of code is very easy to read! 🤩

src/main/java/echo/task/ToDos.java
C5hives: You can consider renaming the `ToDos` class (and every other instance of it) to be singular as it represents only 1 task. It would then adhere to the [coding standard](https://se-education.org/guides/conventions/java/index.html#naming), which states that:
- Plural form should be used on names representing a collection of objects.

src/main/java/echo/tasklist/TaskList.java
C5hives, line 15: Perhaps a more intuitive variable name here? You can consider renaming `taskList` to `tasks` to indicate that it is a collection of `Task` objects.

src/main/java/echo/tasklist/TaskList.java
C5hives: Could this logic be extracted out into the `Task` class directly? This could help with the abstraction of the `Task` class and hide its internal implementation details.

C5hives (commented):
The use of OOP has made the project easy to follow and read. Nice job!
Ignore my previous PR review—I had indicated the wrong review category.

PR #139, opened 2024-08-25
[InfinityTwo] iP

InfinityTwo:
_Features_

- GUI
- ...and that's about it really, nothing special here

**How to use?**

1. Go to the [Github page](https://github.com/InfinityTwo/ip/) of my fork
2. Download the latest release
3. Run the file 😄 

**Which file is the main file for developing and how do I run it?**
Find the file under "src/main/java/infinity/ui/Launcher.java" and use this constructor.
```java
public class Launcher {
        Application.launch(GraphicalUserInterface.class, args);
    }
}
```

Checklist for what's left

- [x] It works
- [x] Test cases updated
- [ ] ???


src/main/java/infinity/command/Command.java
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * Finds the command to execute given the user input and executes it.
+     *
+     * @param currentInput User input.
+     * @param botTasks Tasklist of tasks.
+     * @return The bot output.
+     * @throws InfinityException If there are errors encountered parsing or executing commands.
+     */
+    public static String findCommand(String currentInput, TaskList botTasks) throws InfinityException {
```
chiangqinkang: I feel that this could be clearer if a switch case statement was used instead. Also consider having a secondary function to help with bounds checking, this would better help with information hiding and lead to cleaner code.

src/main/java/infinity/parser/Parser.java
```diff
@@ -0,0 +1,102 @@
...
+     * Also checks if the value is parsable to an integer.
+     *
+     * @param value The value to be checked as a String.
+     * @param min The minimum length of the value.
+     * @param max The maximum length of the value.
+     * @param minNo The minimum value of the number.
+     * @param maxNo The maximum value of the number.
+     * @return True if the value is within the specified range and is a number, false otherwise.
+     */
+    private static boolean isNumberInRange(String value, int min, int max, int minNo, int maxNo) {
```
chiangqinkang, line 38: Perhaps you could split this into multiple lines for clarity, rather than having one long return statement?

chiangqinkang (commented):
Good, comprehensive test case coverage, even added comments for private methods (helpful for understanding code). Could clean up certain parts for better code legibility.

src/main/java/infinity/Infinity.java
```diff
@@ -0,0 +1,58 @@
+package infinity;
+
+import infinity.command.Command;
+import infinity.infinityexception.InfinityException;
+import infinity.storage.Storage;
+import infinity.tasklist.TaskList;
+import infinity.ui.Ui;
+import infinity.ui.components.DialogBox;
+import javafx.scene.image.Image;
+import javafx.scene.layout.VBox;
```
zlch, line 10: I like how the packages are separated nicely.

src/main/java/infinity/command/Command.java
```diff
@@ -0,0 +1,100 @@
...
+        } else if (currentInput.startsWith(KnownCommands.FIND.toString().toLowerCase())
+                && currentInput.length() > 5) {
+
+            botOutput = botTasks.findTasks(currentInput.substring(5));
+
+        } else if (currentInput.startsWith(KnownCommands.SAVE.toString().toLowerCase())) {
+
+            botOutput = Storage.saveFile(botTasks.getTasks());
+
+        } else if (currentInput.equals(KnownCommands.HELP.toString().toLowerCase())) {
```
zlch: Could the toString() and toLowercase() be combined to a single function to increase readability.

zlch (commented):
LGTM. Just some nits.

PR #153, opened 2024-08-25
[valhrd] iP

valhrd:
# Project Yapper 🗣️🗣️🗣️

> "10/10 yapping session"

Yapper is designed to help you stay organized and efficient, all while reducing the mental load of repetitive tasks.

### Why Yapper?
- **Quick Setup**: Get started in minutes.
- **User-Friendly**: Designed with simplicity in mind.
- **Performance**: Optimized for speed and reliability.

### Steps to Get Started
1. Download the JAR file from [here](https://github.com/valhrd/ip/releases/tag/A-Jar).
2. Open the terminal and go to directory where the file is located.
3. Execute `java -jar Yapper.jar`.
4. Start using!

### Key Features
- [ ] Task management
- [ ] Event scheduling
- [ ] Convenient task deletion
- [ ] Real time saving

If you want to see the source code just click [here](https://github.com/valhrd/ip) and run from the Launcher class:

```java
public class Launcher {
    }
}

### CODE REVIEW

File: src/main/java/yapper/app/Deadline.java [hunk 1]
```diff
@@ -0,0 +1,47 @@
+package yapper.app;
+
+import yapper.exceptions.YapperException;
+
+/**
+ * Represents a deadline task in the Yapper chatbot application.
+ */
+public class Deadline extends Task {
+    private String by;
```
waihongteh: It might be better to use a noun here? Maybe something like deadline.

src/main/java/yapper/app/Deadline.java
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * Constructs a new Deadline task with the specified description and deadline.
+     *
+     * @param desc the description of the deadline task
+     * @param by   the date and/or time by which the task must be completed
+     * @throws YapperException if the deadline is empty
+     */
+    public Deadline(String desc, String by) {
```
waihongteh: I personally will prefer description instead of desc as it is more intuitive, but desc looks good still!

src/main/java/yapper/app/DialogBox.java
```diff
@@ -0,0 +1,60 @@
...
+        getChildren().setAll(tmp);
+        setAlignment(Pos.TOP_LEFT);
+    }
+
+    public static DialogBox getUserDialog(String text, Image img) {
+        return new DialogBox(text, img);
+    }
+
+    public static DialogBox getDukeDialog(String text, Image img) {
+        var db = new DialogBox(text, img);
```
waihongteh: db does not sound very understandable, maybe dialogBox would be better?

src/main/java/yapper/app/MainWindow.java
```diff
@@ -0,0 +1,63 @@
...
+    private VBox dialogContainer;
+    @FXML
+    private TextField userInput;
+    @FXML
+    private Button sendButton;
+
+    private Yapper yapper;
+
+    private Image userImage = new Image(this.getClass().getResourceAsStream("/images/you.png"));
+    private Image dukeImage = new Image(this.getClass().getResourceAsStream("/images/icon.png"));
```
waihongteh: Did you forget to change this to yapperImage?

src/main/java/yapper/app/Task.java
```diff
@@ -0,0 +1,126 @@
...
+                    DateTimeFormatter formatter = DateTimeFormatter.ofPattern(fm.toString());
+                    LocalDateTime formattedDateTime = LocalDateTime.parse(time, formatter);
+                    return formattedDateTime.format(DateTimeFormatter.ofPattern("MMM d yyyy hh:mm a"));
+                } catch (DateTimeParseException e) {
+                    continue;
+                }
+            }
+        }
+
+        for (String fm : timeFormats) {
```
waihongteh: Maybe format is more preferrable than fm?

src/main/java/yapper/app/TaskList.java
```diff
@@ -0,0 +1,180 @@
...
+        writeToFile();
+        return Ui.wrapText(texts);
+    }
+
+    /**
+     * Returns the appropriate plural form of the word "task" based on the number of tasks in the list.
+     *
+     * @return a string representing "task" or "tasks"
+     */
+    public String taskPlural() {
```
waihongteh: Since it is a method that converts task to plural maybe verb is better than noun here as the name

src/main/java/yapper/app/Ui.java
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * Constructs a Ui object.
+     */
+    public Ui() {
+    }
+
+    /**
+     * Displays the introduction message and prompts the user.
+     */
+    public static String intro() {
```
waihongteh: introduceMyself() might be better here as it is a method.

waihongteh (commented):
Goodjob overall! Naming looks mostly fine to me.

src/main/java/yapper/app/Deadline.java
valhrd: Will change field name to "deadline" to make things clearer

src/main/java/yapper/app/Deadline.java
valhrd: Will change field name to "description"

src/main/java/yapper/app/DialogBox.java
valhrd: The scope of "db" does not exceed the few lines of code it's used in, so this should not be an issue, although in general more descriptive variable names are better.

If you'd like, you can also refer to the section regarding scratch variables in the coding standard:
https://se-education.org/guides/conventions/java/intermediate.html

src/main/java/yapper/app/MainWindow.java
valhrd: Yes, woops. Will change.

src/main/java/yapper/app/Task.java
valhrd: Will change, thanks.

src/main/java/yapper/app/TaskList.java
valhrd: Will change to a more descriptive name.

src/main/java/yapper/app/Ui.java
valhrd: Yes, it should be more descriptive, thank you for the suggestion.

PR #162, opened 2024-08-25
[Yatoo2] iP

jtooya:
# Tayoo

> "For every minute spent in organizing, an hour is earned." - Benjamin Franklin

Yatoo frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- ~~FAST~~ *SUPER* FAST to use

All you need to do is,

1. download it from [here](https://github.com/jtooya/ip/releases)
2. double-click it
3. add your tasks
4. let it manage your tasks for you :zany_face:

And it is **FREE**

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)

---

```
public class Main {
    }
}
```


src/main/java/tayoo/Parser.java
```diff
@@ -0,0 +1,257 @@
...
+     * converted to upper case. This means that if implementing any new commands, they should be in all upper case when
+     * being added to the if-else chain.
+     *
+     * @param command The user's input in a String format. The command should start with a substring representing the
+     *                type of command,
+     * @return An object that is a subclass of the Command class.
+     * @throws ParserException if there is an error when parsing a command, either because the command is not supported,
+     * or the format of the command is wrong
+     */
+    public static Command parseCommand(String command) throws ParserException {
```
yuvrajaryan: I think that you can consider naming the method like a verb so commandParser instead of parseCommand

src/main/java/tayoo/Storage.java
```diff
@@ -0,0 +1,199 @@
...
+            String line;
+
+            while ((line = reader.readLine()) != null) {
+                lines.add(line);
+            }
+            reader.close();
+        } catch (FileNotFoundException e) {
+            throw new UpdateTxtException("Cannot find tasklist.txt");
+        } catch (IOException e) {
+            throw new UpdateTxtException("An error occurred while reading the file");
```
yuvrajaryan: I like that your names for exceptions are detailed

src/main/java/tayoo/Storage.java
```diff
@@ -0,0 +1,199 @@
...
+
+    /**
+     * Deletes the plaintext representation of a task from the tasklist.txt file. The task to be deleted is identified
+     * by its zero-based index in the tasklist.txt. The method reads the entire file into memory, deletes the relevant
+     * line, then writes the updated content back into the file.
+     *
+     * @param taskNumber the zero-based index of the task that is to be deleted within the tasklist.txt.
+     * @throws DeleteTxtException if tasklist.txt file could not be found, read or updated successfully.
+     */
+    public void deleteTxt(int taskNumber) throws DeleteTxtException {
```
yuvrajaryan: I think that it would be better to spell out text instead of using txt

src/main/java/tayoo/command/AddTaskCommand.java
```diff
@@ -0,0 +1,45 @@
...
+
+import tayoo.exception.TayooException;
+import tayoo.Storage;
+import tayoo.Tasklist;
+import tayoo.Ui;
+import tayoo.tasks.Task;
+
+public class AddTaskCommand extends Command {
+
+    private final Task taskToAdd;
```
yuvrajaryan, line 15: I like the descriptive name for the variable

yuvrajaryan (commented):
LGTM

PR #165, opened 2024-08-25
[lm-44] iP

ty-3401:
<h1>Cook</h1>

> "Time you enjoy wasting is not wasted time." - Marthe Troly-Curtin ([sauce](https://quoteinvestigator.com/2010/06/11/time-you-enjoy/))

The Cook cooks your tasks and serves them in a better format.
---
It is...

- easy to learn
- text-based
- fast to use
- better than your brain at remembering

---

All you need to do is:

1. [DOWNLOAD](https://github.com/lm-44/ip/releases/tag/A-Jar) NOW!!?!
2. double-click it.
3. realise v0.1 doesn't work  :sob:
4. resort to other means of remembering

It is **NOT FREE**, as it takes up more time than it saves.

What can it do?

- [x] Waste your time

<h1>DISCLAIMER:</h1>
<sup><s>v0.1 doesn't work</s> it still doesn't work</sup>

---

If you are a Java programmer, you can use it to learn what not to do. Here's the main method:
```java
public class Cook {
          new Cook(new File("data", "tasks.txt")).run();
      }
}
```


src/main/java/cook/Cook.java
```diff
@@ -0,0 +1,154 @@
...
+import tasks.Deadline;
+import tasks.Event;
+import tasks.ToDo;
+
+/**
+ * Cook class to store main logic and program.
+ */
+public class Cook {
+    private Storage storage;
+    private TaskList tasks;
```
AgentHagu, line 16: I like how you used plural form for your TaskList object since it's a collection of Tasks, good job!

src/main/java/cook/Cook.java
```diff
@@ -0,0 +1,154 @@
...
+                int taskNo;
+
+                try {
+                    taskNo = Integer.parseInt(argumentsHashMap.get(command));
+                } catch (NumberFormatException e) {
+                    this.ui.say("A task must be selected.");
+                    continue;
+                }
+
+                boolean toMark = command.equals("mark");
```
AgentHagu: Perhaps you could rename this boolean `toMark` to use another prefix to better follow the coding standard?

src/main/java/cook/Parser.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    /**
+     * Constructor for Parser class.
+     *
+     * @param input Input to format.
+     * @return HashMap of commands and their corresponding inputs.
+     * @throws InvalidCommandException If command is not recognized.
+     */
+    public HashMap<String, String> readInput(String input) throws InvalidCommandException {
```
AgentHagu: I like the use of a verb for this method `readInput`, improving the code quality. Good job!

src/main/java/cook/Storage.java
```diff
@@ -0,0 +1,42 @@
+package cook;
+
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+
+/**
+ * Storage class to store file locally.
+ */
+public class Storage {
```
AgentHagu, line 13: I like the use of a noun for the class, improving the code quality. Good job!

src/main/java/cook/Cook.java
```diff
@@ -0,0 +1,154 @@
...
+                continue;
+            }
+
+            String command = argumentsHashMap.get("command");
+            if (command.equals("bye")) {
+                return;
+            } else if (command.equals("list")) {
+                this.ui.say(this.tasks.toString());
+            } else if (command.contains("mark") || command.equals("delete")) {
+                int taskNo;
```
AgentHagu: While in general people will understand "no" as "number", perhaps you could use the longer form of the word to avoid possible confusion and improve code quality?

src/test/java/cook/ParserTest.java
```diff
@@ -0,0 +1,206 @@
...
+        validCommandArgs.put("delete", 2);
+        validCommandArgs.put("todo", 2);
+        validCommandArgs.put("deadline", 4);
+        validCommandArgs.put("event", 6);
+
+        return validCommandArgs;
+    }
+
+    @Test
+    public void byeTest() {
```
AgentHagu: Perhaps the test method names could be improved to better follow the three part format in the coding standard?

AgentHagu (commented):
Overall, good job with following the coding standard and code quality guidelines. You used nouns and verbs for classes/variables and methods respectively.

There are only a few nitpicks to fix.

src/test/java/cook/ParserTest.java
```diff
@@ -0,0 +1,206 @@
+package cook;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.HashMap;
+
+import org.junit.jupiter.api.Test;
```
lm-44, line 7: I think this import statement should be moved to line 4 (or line 3 to line 8), to maintain the import statement consistency. https://se-education.org/guides/conventions/java/intermediate.html#statements

src/main/java/cook/Cook.java
ty-3401: 🫡

src/main/java/cook/Cook.java
ty-3401: 🫡

src/test/java/cook/ParserTest.java
ty-3401: 🫡

src/main/java/commands/ToDoCommand.java
```diff
@@ -0,0 +1,46 @@
...
+    public ToDoCommand(String description) throws InvalidInputException {
+        super("todo");
+        this.description = description;
+    }
+
+    /**
+     * Adds ToDo task and saves file.
+     */
+    @Override
+    public String execute(TaskList tasks, Ui ui, Storage storage) {
```
rayray39: For this execute method that is common for ToDoCommand, EventCommand and DeadlineCommand, perhaps you could consider putting it in Command class instead so that you don't have to repeat yourself. But great job on implementing this!

ty-3401: Ah! The main reason I separated them out was because each command would need to create a specific instance of a different class.
``` java
// Create ToDo instance
new Todo(this.description)
// Create Deadline instance
new Deadline(this.description, this.by)
// Create Event instance
new Event(this.description, this.from, this.to)
```
I don't know of a clean way to combine all of this into a single method in the command class since it only contains the description attribute. Even if I did, the method would be inherited by the other commands which I don't really want to happen since they don't even make use of it.

The other option I thought of was to create another abstract class inheriting from Command, which all task creating commands will inherit from. This will reduce code duplication by containing it within the new abstract class, but I'm unsure of the feasibility of calling a super method within its overridden method in Java. Do let me know if you have other thoughts on this.

PR #166, opened 2024-08-26
[fymbc] iP

fymbc:
## Friday

> “Why is Gamora?” – Drax ([source](https://www.youtube.com/watch?v=dQw4w9WgXcQ))

Friday ensures you get all things done before Friday!. It's,

- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,

1. download it from [here](https://github.com/fymbc/ip).
2. run the jar file.
3. add your tasks.

And it is **FREE**!

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)

```java
public class Main {
    }
}
```


src/test/java/tasks/TodoTest.java
```diff
@@ -0,0 +1,40 @@
+package tasks;
+
+import org.junit.jupiter.api.Test;
+import java.time.LocalDate;
+
+import static org.junit.jupiter.api.Assertions.*;
```
wkxcass, line 3: Perhaps listing out the imported classes explicitly? I noticed similar issues in several other places too!

src/main/java/tasks/Task.java
```diff
@@ -0,0 +1,34 @@
...
+
+    public void setDone() {
+        this.done = true;
+    }
+
+    public void setUndone() {
+        this.done = false;
+    }
+
+    public String getName() { return this.name;
```
wkxcass: Perhaps moving the `return` statement into another line below?

src/main/java/storage/TaskList.java
```diff
@@ -0,0 +1,138 @@
...
+            } else {
+                return str.trim();
+            }
+        }
+    }
+
+    public LocalDate getDate(String type, Scanner sc) {
+        try {
+            switch (type) {
+                case "d":
```
wkxcass: Perhaps removing the indentations for `case` clauses (as per the course's coding standards)? I noticed the same issue in several other places too!

src/main/java/commands/removeCommand.java
```diff
@@ -0,0 +1,26 @@
+package commands;
+import storage.Storage;
```
wkxcass: Perhaps adding a line between `package commands;` and `import storage.Storage;`? I found the same issue in several other places too!

src/main/java/Friday.java
```diff
@@ -0,0 +1,41 @@
+import commands.Command;
```
wkxcass: Perhaps putting the class into a package (as per the course's coding standards)? I would suggest creating a `friday` package, that would contain the `Friday` class as well as the other packages!

wkxcass (commented):
Overall, I like how readable your code is! 👍
Just a couple of nits to fix.

src/main/java/Friday.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * The main class that runs the Friday application, a to-do list chatbot.
+ * The Friday class initializes the necessary components (Storage, TaskList, and UI)
+ * and manages the application's main loop, handling user input and executing commands.
+ */
+public class Friday {
+
+    private final Storage storage;
+    private final TaskList master;
```
cth06-Github: May I check if there is any reason why "master" is used as the naming? Would "tasks" be a better naming as it conveys the meaning that it is about tasks?

src/main/java/commands/AddCommand.java
```diff
@@ -0,0 +1,40 @@
...
+import storage.TaskList;
+import ui.UI;
+
+/**
+ * Represents a command to add a new task to the task list.
+ * The AddCommand class implements the Command interface and handles the addition of a task.
+ */
+public class AddCommand implements Command {
+
+    private final String desc;
```
cth06-Github: I believe that "desc" was written as a shortform of "description", but I think the shortform may be slightly ambigious, would keeping the full form be better? I think "description" is not a long word but I concede that this is a subjective matter!

src/main/java/commands/AddCommand.java
```diff
@@ -0,0 +1,40 @@
...
+     * Executes the add task command, adding a new task to the task list.
+     * The task list is then saved to storage, and a line separator is printed.
+     *
+     * @param storage the Storage object for handling task persistence
+     * @param master the TaskList object containing the list of tasks
+     * @param ui the UI object for interacting with the user
+     * @return false, indicating that the application should not terminate
+     */
+    @Override
+    public boolean execute(Storage storage, TaskList master, UI ui) {
```
cth06-Github: May I check what this function does? My understanding is that the function manipulates the TaskList master and saves the list and print the line during execution, but after which it also returns whether it is executed. Would it be better to separate the functions into 2? We can have one `void execute()` function that does all the manipulation, and another boolean function to indicate whether the function is executed. This is to follow Single Responsibility Principle.

I notice that this function always return false, thus could I check what does this boolean status represent? I think The boolean function should ideally be named as `isExecuted()` or `hasStatus()` - with the prefix "is" or "has" followed by whatever the boolean status represent.

Same comment for `execute()` in other parts of the code.

src/main/java/commands/FindCommand.java
```diff
@@ -0,0 +1,52 @@
...
+     * @param master the TaskList object containing the list of tasks
+     * @param ui the UI object for interacting with the user
+     * @return false, indicating that the application should not terminate
+     */
+    @Override
+    public boolean execute(Storage storage, TaskList master, UI ui) {
+        String s = "";
+        for (int i = 0; i < master.getSize(); i++) {
+            Task task = master.getParent().get(i);
+            if (task.containsWord(this.desc)) {
```
cth06-Github: Should this be `task.containWord(this.desc)`? (The verb "contains" should be in the base or root form.)

src/main/java/commands/MarkCommand.java
```diff
@@ -0,0 +1,48 @@
...
+     * @param master the TaskList object containing the list of tasks
+     * @param ui the UI object for interacting with the user
+     * @return false, indicating that the application should not terminate
+     */
+    @Override
+    public boolean execute(Storage storage, TaskList master, UI ui) {
+        try {
+            int index = Integer.parseInt(this.desc);
+
+            master.doneTask(this.action, index - 1);
```
cth06-Github: May I clarify if `doneTask()` mean `markTaskAsDone()` or `showDoneTask()` (I.e. `showTaskDone()` -- show task that was done?

src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,55 @@
...
+        }
+        case "mark", "unmark" -> {
+            if (parsed.length > 1) {
+                yield new MarkCommand(parsed[0], parsed[1]);
+            }
+            yield new InvalidCommand(parsed[0]);
+        }
+        case "help" -> new HelpCommand();
+        default -> new InvalidCommand(parsed[0]);
+        };
```
cth06-Github: I like how the arrow label way of writing switch - case block is used.

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,161 @@
...
+     *
+     * @param line the line from the file representing a task
+     * @return the Task object created from the line
+     * @throws IOException if the line format is incorrect
+     */
+    public Task parseTask(String line) throws IOException {
+        String[] values = line.split(",");
+        Task task;
+        switch (values[0]) {
+        case "td":
```
cth06-Github, line 86: Would it be better if a comment is written to explain that case "td" means an instruction related to ToDo task? Same goes to "d" and "e".

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,161 @@
...
+                }
+                if (task instanceof Event e) {
+                    str += "e," + e.getName();
+                    if (task.isDone()) {
+                        str += ",y,";
+                    } else {
+                        str += ",n,";
+                    }
+                    str += e.getStart() + "," + e.getEnd();
+                }
```
cth06-Github: There seem to be repeated lines of code. If these functions are written as part of the Todo, Deadline, Event class using the same name, then we can make use of polymorphism and the output of the string to be saved into Storage will automatically change without the need for `instanceof`. It will be beneficial because if we need to create another new class that inherits from Task, we do not have to add in 
```Java
if (instanceof newClass n) {
    // code on how the string should be printed/saved into the text file.
} 
``` 
to give an example.

Definitely feel free to disagree because code design wise...while I think this suggestion will make the code better, maybe there are problems I did not notice such that "this functionality should not be written in this class but another class".

src/main/java/storage/TaskList.java
```diff
@@ -0,0 +1,204 @@
...
+import java.time.format.DateTimeParseException;
+import java.util.ArrayList;
+import java.util.Scanner;
+
+/**
+ * Manages a list of tasks, including adding, removing, and marking tasks as done or undone.
+ * The TaskList class is responsible for handling all operations on the tasks stored in the list.
+ */
+public class TaskList {
+    private final ArrayList<Task> parent;
```
cth06-Github: Would it be better if the variable is named as tasks? (or anything else that is plural?) I am guessing that "parent" was used to indicate that under the hood, or, at its most fundamental level, the TaskList class is like an `ArrayList<Task>` ? So it does feel like `ArrayList<Task>` is the parent? I guess TaskList and ArrayList are associated but they aren't parent-child relationship as if it was inheritance, so it might be a misnomer. Moreover I think it's more important to show what does this attribute do / mean, rather than showing the relationship or association, If my speculation on why the word "parent" is used is true.

src/main/java/storage/TaskList.java
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    /**
+     * Marks a task as done or undone based on the specified action.
+     * The user can choose to mark a task as done or undone, and a confirmation message is displayed.
+     *
+     * @param action the action to perform ("mark" for done, any other value for undone)
+     * @param task the index of the task to be marked
+     */
+    public void doneTask(String action, int task) {
```
cth06-Github: Would `markTaskAsDone()` be better, although the name is longer? In one of the above comments, I asked whether it should be `markTaskAsDone()` or `showDoneTask()` as I feel that it is not exactly clear whether "done" in this case is meant to be a verb but written in wrong tense or meant to describe the task (adjective).

src/main/java/storage/TaskList.java
```diff
@@ -0,0 +1,204 @@
...
+     * @param type the type of date expected ("d" for deadline, "es" for event start, "ee" for event end)
+     * @param sc the Scanner object used to read user input
+     * @return the user's input as a LocalDate object
+     */
+    public LocalDate getDate(String type, Scanner sc) {
+        try {
+            switch (type) {
+            case "d":
+                return LocalDate.parse(getInputFromUser(sc, "Deadline (in dd mm yyyy) > "), this.inputFormatter);
+            case "es":
```
cth06-Github: What does "es" stand for?

src/main/java/storage/TaskList.java
```diff
@@ -0,0 +1,204 @@
...
+                return LocalDate.parse(getInputFromUser(sc, "Start Date (in dd mm yyyy) > "), this.inputFormatter);
+            case "ee":
```
cth06-Github: What does "ee" stand for?

src/main/java/storage/TaskList.java
```diff
@@ -0,0 +1,204 @@
...
+        return getDate(type, sc);
+    }
+
+    /**
+     * Counts the number of tasks that are either completed or incomplete.
+     *
+     * @param test true to count completed tasks, false to count incomplete tasks
+     * @return the number of tasks that match the specified completion status
+     */
+    public int countCompleted(boolean test) {
```
cth06-Github:
```suggestion
    public int countCompleted(boolean isTested) {
```

cth06-Github: Would it be possible to clarify what boolean test variable represents? I'm not sure whether my interpretation is correct.

src/main/java/tasks/Event.java
```diff
@@ -0,0 +1,58 @@
...
+        return String.format("%s | Event from %s to %s", super.toString(),
+                outputFormatter.format(this.startDate), outputFormatter.format(this.endDate));
+    }
+
+    /**
+     * Returns the start date of the event.
+     *
+     * @return the start date of the event
+     */
+    public LocalDate getStart() {
```
cth06-Github:
```suggestion
    public LocalDate getStartDate() {
```

src/main/java/tasks/Event.java
```diff
@@ -0,0 +1,58 @@
...
+        return startDate;
+    }
+
+    /**
+     * Returns the end date of the event.
+     *
+     * @return the end date of the event
+     */
+    public LocalDate getEnd() {
```
cth06-Github:
```suggestion
    public LocalDate getEndDate() {
```

src/main/java/tasks/Task.java
```diff
@@ -0,0 +1,71 @@
+package tasks;
+
+/**
+ * Represents an abstract task with a name and a completion status.
+ * This class provides common functionality for various types of tasks.
+ * The task is either marked as done or not done.
+ */
+public abstract class Task {
+    private final String name;
+    private boolean done;
```
cth06-Github:
```suggestion
    private boolean isDone;
```

cth06-Github (commented):
Good work, code naming mostly adhered to! I hope the review comments are alright!

src/main/java/tasks/Task.java
fymbc: Agreed! Thank you!

src/main/java/storage/TaskList.java
fymbc: Agreed! Thank you!

src/main/java/commands/removeCommand.java
fymbc: Agreed! Thank you!

src/main/java/Friday.java
fymbc: Agreed! Thank you!

src/test/java/tasks/TodoTest.java
fymbc: Hmm, I was following the notes which I believe imported it as a wildcard. I'll see how I can better list this. Thank you!

src/main/java/tasks/Task.java
fymbc: Thank you for catching this!

src/main/java/tasks/Event.java
fymbc: Thank you for catching this!

src/main/java/tasks/Event.java
fymbc: Thank you for catching this!

src/main/java/storage/TaskList.java
fymbc: It's supposed to be a dual use function, where it returns the number of completed tasks if boolean is set to true, and number of uncompleted tasks if boolean is set to false. But I see where the confusion may be, have made amendments. Thank you!

src/main/java/storage/TaskList.java
fymbc: Event end! Which denotes the date expected from the input should be the event end date.

src/main/java/commands/MarkCommand.java
fymbc: The former! Understand where the confusion may be. Have changed the naming. Thanks for pointing it out!

src/main/java/commands/FindCommand.java
fymbc: Thanks for catching this! Have amended.

src/main/java/commands/AddCommand.java
fymbc: Agreed! Have amended this for all commands. Thank you!

src/main/java/Friday.java
fymbc: No particular reason, just the first variable that came to find. Agreed it may be ambiguous. Have amended. Thank you!

src/main/java/commands/AddCommand.java
fymbc: The function was augmented to return a boolean function to distinguish between an exit command (which will return true) versus all other commands (which will return false), so that the program can catch and terminate when "bye" is inputted by the user. The aim is not to determine if command has execute successfully.

Noted on the Single Responsibility Principle. Let me try to think how I can improve this. Thank you!

src/main/java/storage/Storage.java
fymbc: Agreed! Have added further explanation in the JavaDocs. Thank you!

src/main/java/storage/Storage.java
fymbc: You have a point. The motivation is to parse it into a CSV-esque format that reduces potential errors in parsing.

Perhaps I can shift this "to-CSV-format" functionality into the tasks itself and make use of polymorphism.

Thank you!

src/main/java/storage/TaskList.java
fymbc: Agreed! Amended! Thank you!

src/main/java/storage/TaskList.java
fymbc: I agree! Noted and amended. Thanks for catching it!

PR #180, opened 2024-08-26
[leslieyip02] iP

leslieyip02:
# Alice

Alice is a generic name that is commonly used when discussing cryptography. ([relevant xkcd](https://xkcd.com/1323/))

> Alice and Bob want to exchange a message or cryptographic key. Alice wants to send a private message to Bob, ...

## Features
- [x] Adding tasks
- [x] Removing tasks
- [x] Marking tasks as complete/incomplete
- [x] Listing tasks
- [x]  Saving tasks
- [ ] Interactive GUI

## Changes
- Rename chatbot
- Add `command`, `parser`, `storage`, `task` and `alice` packages

### Types of Tasks
1. To-Dos 👍
2. Deadlines 🎯
3. Events 📆

## Setup

- Download the `.jar` file [here](https://github.com/leslieyip02/ip/releases/tag/A-Jar)

OR

- Clone the repository and run `Alice::main`:

```
public class Alice {
    ...

        Alice alice = new Alice();

        alice.greet();
        alice.listen();
        alice.bye();
    }
}
```


src/main/java/alice/Alice.java
```diff
@@ -0,0 +1,65 @@
...
+import alice.ui.Ui;
+
+public class Alice {
+    private static final String NAME = "Alice";
+    private static final String DATA_DIRECTORY_PATH = "./data";
+    private static final String TASKS_FILE_NAME = "tasks.jsonl";
+
+    private final Ui ui;
+    private final Storage storage;
+    private final TaskList taskManager;
```
somethingfishyfishy: surprising that the naming for this is taskManager

Does this mean that the TaskList class is doing more than managing a List<Task> Object?

src/main/java/alice/Alice.java
```diff
@@ -0,0 +1,65 @@
...
+    private void bye() {
+        ui.say("Bye. Hope to see you again soon!");
+    }
+
+    private void echo(String line) {
+        // echo user inputs
+        ui.say(String.format("%s", line));
+    }
+
+    private void listen() {
```
somethingfishyfishy: Perhaps a better naming for this method would be run(), or loop()

Since you are not only listening for the command, but executing the command when it is not 'bye'

src/main/java/alice/ui/Ui.java
```diff
@@ -0,0 +1,31 @@
+package alice.ui;
+
+/** Handles output to the user. */
+public class Ui {
```
somethingfishyfishy: I noticed that you always create a new Ui Object and store it as an attribute for other classes. I'm not saying it is wrong.

But you could make the below methods static methods so you don't require a Ui object to say things.

src/main/java/alice/command/Command.java
```diff
@@ -0,0 +1,75 @@
...
+     * @param  ui       the ui instance
+     * @param  taskList the taskList instance
+     * @return          the parsed command
+     */
+    public static Command fromInput(String line, Ui ui, TaskList taskList) throws IllegalArgumentException {
+        String input = line.toUpperCase();
+        if (input.startsWith(Action.TODO.name()) ||
+            input.startsWith(Action.DEADLINE.name()) ||
+            input.startsWith(Action.EVENT.name())) {
+            return new AddTask(ui, taskList);
```
somethingfishyfishy: I would think that AddTask (and the related) sounds more like a method name than a class name.

But there is nothing wrong with this currently.

src/main/java/alice/parser/DateParser.java
```diff
@@ -0,0 +1,45 @@
...
+     * @param  dateString             the date string to parse
+     * @throws DateTimeParseException if the date string is not well=formed
+     * @return                        the parsed LocalDateTime
+     */
+    public static LocalDateTime parseDateString(String dateString) throws DateTimeParseException {
+        for (DateTimeFormatter formatter :  DATETIME_FORMATTERS) {
+            try {
+                return LocalDateTime.parse(dateString, formatter);
+            } catch (DateTimeParseException exception) {
+                // try another format
```
somethingfishyfishy, line 29: perhaps print something, even if you're not doing anything with the exception object caught

src/main/java/alice/task/InvalidTaskException.java
```diff
@@ -0,0 +1,11 @@
+package alice.task;
+
+public class InvalidTaskException extends Exception {
```
somethingfishyfishy, line 4: any javadoc comment?

src/main/java/alice/command/DeleteTask.java
```diff
@@ -0,0 +1,47 @@
...
+        if (tokens.length != 2) {
+            throw new InvalidTaskException("Missing task number. Usage: delete <task number>");
+        }
+
+        int taskNumber;
+        try {
+            taskNumber = Integer.parseInt(tokens[1]);
+        } catch (NumberFormatException exception) {
+            throw new InvalidTaskException("Invalid task number. Usage: delete <task number>");
+        }
```
yugan01, line 34: Should this be extracted out? As i observe the same code being repeated for DeleteTask/MarkTask and UnMarkTask

src/main/java/alice/parser/DateParser.java
```diff
@@ -0,0 +1,45 @@
...
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+/** Handles datetime and string conversions. */
+public class DateParser {
+    private static final DateTimeFormatter[] DATETIME_FORMATTERS = new DateTimeFormatter[]{
+        DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm"),
+        DateTimeFormatter.ofPattern("yyyy-MM-dd HHmm"),
+        DateTimeFormatter.ofPattern("dd-MM-yyyy HH:mm"),
+        DateTimeFormatter.ofPattern("dd-MM-yyyy HHmm"),
```
yugan01, line 13: I like how you have a separate class to parse dates of varying formats.

src/main/java/alice/storage/TaskList.java
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+     * Adds task to list and saves to disk.
+     *
+     * @param  task        the task to add
+     * @throws IOException if the tasks cannot be saved
+     */
+    public void addTask(Task task) throws IOException {
+        tasks.add(task);
+        saveTasks();
```
yugan01: I understand that saveTasks() is called whenever a task is modified (added, deleted, marked). Is there any reason why saveTasks is not called at the end, when user says goodbye?

yugan01 (commented):
LGTM, your code was easy to follow and understand. I noticed a few situations where abstraction could be beneficial.

PR #207, opened 2024-08-26
[Nicholas Koh Zi Lun] iP

Nicholaskoh1:
# DukePro
> "Your mind is for having ideas, not holding them." - David Allen ([source](https://dansilvestre.com/productivity-quotes/))

DukePro frees your mind of having to remember things you need to do. It's,
- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)

```ruby
public class Main {
    }
}
```


src/main/java/TaskList.java
```diff
@@ -0,0 +1,130 @@
...
+                // Check if the data file is corrupted
+                if (taskParts.length < 3) {
+                    throw new FileCorruptException();
+                }
+
+                String taskType = taskParts[0];
+                boolean isDone = taskParts[1].equals("1");
+                String taskDescription = taskParts[2];
+                Task currentTask = null;
+
```
sandyk0105: Good use of variable names!

src/main/java/jeff/parser/Parser.java
```diff
@@ -0,0 +1,114 @@
...
+            return new DeleteCommand(input);
+        } else if (input.startsWith("todo") || input.startsWith("deadline") || input.startsWith("event")) {
+            return new AddCommand(input);
+        } else if (input.startsWith("task")) {
+            return new DateCommand(input);
+        } else if (input.startsWith("find")) {
+            return new FindCommand(input);
+        } else {
+            throw new JeffException();
+        }
```
C5hives, line 51: Should the logic of parsing the first word in the string be extracted out?

You can consider creating an Enum and letting said Enum handle the logic of parsing the first word into a relevant command. This way, the chain of if-else statements can be replaced with a switch statement for better readability.

C5hives (changes requested):
LGTM—the use of OOP makes the project easy to read and understand. 😄

src/main/java/jeff/exception/JeffException.java
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Returns the message of the error.
+     *
+     * @return the error message
+     */
+    @Override
+    public String toString() {
+        return this.message;
+    }
+}
```
C5hives: I like how you created a custom exception for the chatbot—it'll help with managing current (and future) exceptions in a more precise manner!

src/main/java/jeff/parser/Parser.java
C5hives: Should the logic of parsing the first word in the string be extracted out?

You can consider creating an Enum and letting said Enum handle the logic of parsing the first word into a relevant command. This way, the chain of if-else statements can be replaced with a switch statement for better readability.

src/main/java/jeff/parser/Parser.java
```diff
@@ -0,0 +1,114 @@
...
+        StringBuilder resultText = new StringBuilder();
+
+        // Add a space in front each line
+        for (String line : lines) {
+            resultText.append(" ").append(line).append("\n");
+        }
+
+        // Convert the StringBuilder back to a String
+        return resultText.toString();
+    }
```
C5hives: The method name `prettyText` might not be the best way to name this static method.

1. Although the word 'pretty' can be used as a verb, it may not be the most intuitive interpretation of the word ('pretty' can be an adjective to describe the word 'text' as well).
2. Moreover, the method name also may not be most descriptive in telling readers what the function does, which is to add a leading space to every line in the string.

Perhaps you can consider renaming the method to
- `convertToPrettyText()` (method name that starts with a verb) or
- `toPrettyText()` (similar to the `toString()` method name)

to address issue 1, or you could consider something even more descriptive (and possibly overkill), such as:
- `addLeadingSpaceToText()` (very explicit in terms of what the method does)

to address issue 2.

This ultimately boils down to personal preference, but these would be my suggestions. :)

src/main/java/jeff/task/TaskList.java
```diff
@@ -0,0 +1,195 @@
...
+import java.util.Scanner;
+
+import jeff.exception.FileCorruptException;
+import jeff.exception.JeffException;
+
+/**
+ * Represents a list of tasks.
+ */
+public class TaskList {
+    private final ArrayList<Task> tasks = new ArrayList<>();
```
C5hives, line 18: According to the [coding standard](https://se-education.org/guides/conventions/java/index.html#naming), constant names must be all uppercase using underscore to separate words (aka SCREAMING_SNAKE_CASE).

You may want to consider renaming the variable to screaming snake case, or to remove the `final` keyword (I know the Intellj IDE will suggest you to make this variable `final`).

src/main/java/jeff/task/TaskList.java
```diff
@@ -0,0 +1,195 @@
...
+        try {
+            // Read every line in the scanner
+            while (scanner.hasNext()) {
+                // Get the task information
+                String[] taskParts = scanner.nextLine().split(" \\| ");
+
+                // Check if the data file is corrupted
+                if (taskParts.length < 3) {
+                    throw new FileCorruptException();
+                }
```
C5hives, line 54: I like the input validation sprinkled throughout the project to defend against the myriad of possible bad inputs. Nice job!

src/test/java/jeff/command/AddCommandTest.java
C5hives, line 1: Your tests are very comprehensive. Nice job!

C5hives (commented):
LGTM—the use of OOP makes the project easy to read and understand. 😄
Ignore my previous PR comment—I had indicated the wrong review category.

src/main/java/jeff/parser/Parser.java
ITZXITZX: I agree with @C5hives.

Here is sounds like 'pretty' is used as a adjective to describe 'Text'.

src/main/java/jeff/command/AddCommand.java
```diff
@@ -0,0 +1,122 @@
...
+                String[] deadlineParts = taskDescription.split(" /by ", 2);
+                String deadlineContent = deadlineParts[0];
+                String deadlinePeriod = deadlineParts.length > 1 ? deadlineParts[1] : "";
+
+                // Check if the format is correct
+                try {
+                    targetTask = new DeadlineTask(deadlineContent, Parser.getLocalDateTime(deadlinePeriod));
+                } catch (DateTimeParseException e) {
+                    throw new JeffException(
+                            "The format is wrong! It should be \"deadline xx /by yyyy-mm-dd HH:mm or hh:mm AM/PM\"!");
```
ITZXITZX: Is this line [too long](https://se-education.org/guides/conventions/java/intermediate.html#:~:text=Line%20length%20should%20be%20no%20longer%20than%20120%20chars.)?

src/main/java/jeff/command/AddCommand.java
```diff
@@ -0,0 +1,122 @@
...
+                String eventContent = eventParts[0];
+                String eventPeriod = eventParts.length > 1 ? eventParts[1] : "";
+                String[] eventPeriodParts = eventPeriod.split(" /to ", 2);
+                String startPeriod = eventPeriodParts[0];
+                String endPeriod = eventPeriodParts.length > 1 ? eventPeriodParts[1] : "";
+
+                // Check if the format is correct
+                try {
+                    targetTask = new EventTask(
+                            eventContent, Parser.getLocalDateTime(startPeriod), Parser.getLocalDateTime(endPeriod));
```
ITZXITZX: Will breaking the line here [improve readability](https://se-education.org/guides/conventions/java/intermediate.html#:~:text=Place%20line%20break%20to%20improve%20readability)?

src/main/java/jeff/task/Task.java
```diff
@@ -0,0 +1,122 @@
...
+        return false;
+    }
+
+    /**
+     * Checks if the task description contains the given name.
+     *
+     * @param name Given name to check against.
+     * @return true if the task description contains the given name and false otherwise.
+     */
+    public boolean contains(String name) {
```
ITZXITZX: Is there a clearer name for `contains`? Since its specifically searching within the task description only, and not the other fields.

src/main/java/jeff/task/TaskList.java
```diff
@@ -0,0 +1,195 @@
...
+    public List<Task> filterByName(String name) {
+        return this.tasks.stream().filter(task -> task.contains(name)).toList();
+    }
+
+    /**
+     * Returns a list of the file string representation of the tasks.
+     *
+     * @return A list of the file string representation of the tasks.
+     */
+    public List<String> toFileStrings() {
```
ITZXITZX: Is there a clearer name for `toFileStrings`, since it is returning not a String but a List.

ITZXITZX (commented):
Overall your code seems good, with some minor cosmetic issues.

src/main/java/jeff/parser/Parser.java
Nicholaskoh1: Thanks for pointing it out!

I wanted to prompt the user with the format of the command when they forget to put a spacing after the command word. For example, "findborrow book". If i try to use an Enum and extract the first word, the first word in this case would be "findborrow" instead of "find". This would then prompt a different message instead and thus I didn't want to use Enum for this.

If you have a solution to this, do let me know!

src/main/java/jeff/parser/Parser.java
Nicholaskoh1: Thanks for pointing it out! I like the suggestion for `addLeadingSpaceToText()` so I would be replacing the function name with this.

src/main/java/jeff/command/AddCommand.java
Nicholaskoh1: Thanks for pointing it out! I will be changing it in my next commit.

src/main/java/jeff/command/AddCommand.java
Nicholaskoh1: Yup I agree that I should break the line here to make it easier to read. Thanks for pointing it out!

src/main/java/jeff/task/TaskList.java
Nicholaskoh1: Thanks for pointing it out! I will try to change it to make it clearer.

src/main/java/jeff/task/Task.java
Nicholaskoh1: Thanks for pointing it out! I will change it to make it clearer.

PR #217, opened 2024-08-26
[zackjh] iP

zackjh:
# Evan

> "The biggest constraint of the next 100 years of computing is the idea of metaphors." - Evan Spiegel, CEO of Snap Inc. ([source](https://sg.finance.yahoo.com/news/11-brilliant-quotes-evan-spiegel-191024037.html?guccounter=1))

Evan frees your mind of having to remember things you need to do. It's:

- text-based
- intuitive
- ~~fast~~ really fast

All you need to do is:

1. Download it from [here](https://github.com/zackjh/ip/releases)
2. Double-click it
3. Add your tasks
4. Reap the benefits of using Evan

Best of all, it's **free**! 😁

Features:

- [x] Managing tasks (todos, deadlines, and events)
- [ ] Reminders (coming soon)

---

If you are a Java programmer, you can use it to practise your Java skills too! Here's the `main` method:
```java
    new Evan("data/tasks.txt").run();
}
```


src/main/java/evan/command/DeadlineCommand.java
```diff
@@ -0,0 +1,43 @@
...
+     * @param by          End date/time of the Deadline to be created.
+     */
+    public DeadlineCommand(String description, String by) {
+        this.description = description;
+        this.by = by;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
```
raageth, line 26: I like how using `{@inheritDoc}` keeps the code clean.

src/main/java/evan/command/Command.java
```diff
@@ -0,0 +1,29 @@
...
+     * @param storage  Storage object that saves the chatbot's data to the storage .txt file.
+     * @throws NoSuchTaskException If the command attempts to perform an operation on a Task that doesn't exist.
+     */
+    public abstract void execute(TaskList taskList, Ui ui, Storage storage) throws NoSuchTaskException;
+
+    /**
+     * Returns true if the chatbot should exit the session after execution of the command, and false if the chatbot
+     * should continue to prompt the user for input.
+     */
+    public abstract boolean isExit();
```
raageth: Since `ExitCommand` is the only place where `isExit()` returns `true`,  why not set `isExit()` to return `false` here as a default value?

src/main/java/evan/main/Storage.java
```diff
@@ -0,0 +1,134 @@
...
+        }
+    }
+
+    /**
+     * Returns an ArrayList of Tasks containing tasks decoded from the storage .txt file.
+     *
+     * @return ArrayList of Tasks containing tasks decoded from the storage .txt file.
+     * @throws LoadingException If there is an error reading the storage .txt file.
+     */
+    public ArrayList<Task> load() throws LoadingException {
```
raageth: Could a more specific method name be used in this case?

src/main/java/evan/task/DateTime.java
```diff
@@ -0,0 +1,183 @@
...
+
+    // Method to check if the input is a string
+
+    /**
+     * Checks if the DateTime object is using a STRING.
+     *
+     * @return True if the DateTime object is using a STRING, and false otherwise.
+     */
+    public boolean isString() {
+        return type == InputType.STRING;
```
raageth, line 101: Would using `type.equals(InputType.STRING)` be better?

src/test/java/evan/task/DateTimeTest.java
```diff
@@ -0,0 +1,32 @@
...
+        assertEquals(DateTime.InputType.STRING, DateTime.parseInput("2024-12-32 1261").type);
+
+        // Invalid date, valid time → STRING
+        assertEquals(DateTime.InputType.STRING, DateTime.parseInput("2024-00-25 1200").type);
+
+        // Valid date → DATE
+        assertEquals(DateTime.InputType.DATE, DateTime.parseInput("2024-04-06").type);
+
+        // Valid date and time → DATETIME
+        assertEquals(DateTime.InputType.DATETIME, DateTime.parseInput("2024-09-10 2359").type);
```
raageth, line 29: I like how thorough the testing is.

raageth (commented):
LGTM. The code was well written and I could not find much coding standard issues.

src/main/java/evan/command/Command.java
colinhia: Adding on to the point above, perhaps another approach is instead of a method, you could consider using a boolean field instantiated to be false by default? Then you could just access the field instead of having to implement the method everywhere

src/main/java/evan/command/TodoCommand.java
```diff
@@ -0,0 +1,40 @@
...
+import evan.main.Storage;
+import evan.main.TaskList;
+import evan.main.Ui;
+import evan.task.Todo;
+
+/**
+ * Represents a command that the chatbot can execute to create a Todo.
+ */
+public class TodoCommand extends Command {
+    private final String description;
```
colinhia, line 10: I noticed that the String description is common to the Todo/Deadline/Event commands, perhaps there could be another abstract class grouping these commands for greater encapsulation?

src/main/java/evan/main/Evan.java
```diff
@@ -0,0 +1,72 @@
...
+            } catch (InvalidUserInputException e) {
+                ui.showError(e.getMessage());
+                ui.showValidCommands();
+            } catch (NoSuchTaskException | SavingException e) {
+                ui.showError(e.getMessage());
+            }
+        }
+        System.exit(0);
+    }
+}
```
colinhia: I liked how specific your exceptions were! Perhaps one suggestion I would make is have them grouped under an abstract EvanException class? Then you might be able to handle the exceptions more easily in the main Evan class.

src/main/java/evan/main/Storage.java
```diff
@@ -0,0 +1,134 @@
...
+            switch (taskType) {
+            case "T" -> task = new Todo(description);
+            case "D" -> {
+                String by = parts[3];
+                task = new Deadline(description, by);
+            }
+            case "E" -> {
+                String from = parts[3];
+                String to = parts[4];
+                task = new Event(description, from, to);
```
colinhia: I liked the use of the switch statement here!

src/main/java/evan/task/Task.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    /**
+     * Returns a String representation of the Task used for display purposes.
+     */
+    @Override
+    public String toString() {
+        return String.format("[%s] %s", this.getStatusIcon(), this.description);
+    }
+}
```
colinhia, line 61: I liked the grouping of common methods into the overarching abstract task class!

colinhia (commented):
Overall great job! I liked how consistent the coding style was throughout the project, from the import statements, to the indentation in the code. I think it is no easy feat to maintain that, especially with the use of switch statements and the such. Keep it up!

PR #224, opened 2024-08-26
[Keerthi G Kaarthik] iP

keerthigkaarthik:
Bob
> My best fren - Me

Bob helps you remember the things you need to do so you can focus on actually doing them! He is

- text-based
- based
- QUICK

You just need to

1. download
2. double click
3. add your tasks
4. profit!!

Bob is so kinda and willing to do all of this (and more in the future) for **free**

- [X] Multiple Tasks type
- [X] Deadlines
- [X] Mark as complete
- [ ] Sorting (coming soon)
- [ ] Reminders (coming soon)

```
public class Main {
    }
}
```


src/main/java/Task/TaskList.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /**
+     * Prints all tasks in the list with their indices.
+     */
+    public void printList() {
+        int index = 1;
+        System.out.println();
+        for (Task task : this.tasks) {
+            System.out.printf("     %d. %s\n", index, task);
```
valhrd: Nice use of format specifiers, I never thought of that actually.

src/main/java/UI/Parser.java
```diff
@@ -0,0 +1,142 @@
...
+        if (arguments.isEmpty()) throw new BotException("Please provide a task description.");
+        new ToDo(arguments);
+        UI.taskAddedMsg();
+    }
+
+    private static void handleDeadline(String arguments) throws BotException {
+        int byIndex = arguments.indexOf(" /by ");
+        if (byIndex == -1) throw new BotException("Please format your instructions correctly. E.g., deadline [task] /by [MMddyyyy HHmm]");
+        String taskDescription = arguments.substring(0, byIndex);
+        String deadline = arguments.substring(byIndex + 5);
```
valhrd: Have you considered:
```suggestion
        String deadline = arguments.split(" /by ")[1].trim();
        if (deadline.isEmpty()) {
            throw new BotException("Please format your instructions correctly. E.g., deadline [task] /by [MMddyyyy HHmm]");
        }
```

src/main/java/UI/UI.java
```diff
@@ -0,0 +1,68 @@
...
+    static void taskAddedMsg() {
+        System.out.println("Sure! I've added that in for you.");
+        TaskList.mainTaskList.printList();
+    }
+
+    /**
+     * Displays a separator line to the console.
+     */
+    public void showSeparator() {
+        System.out.println("--------------------------------------------------");
```
valhrd: Can we try setting this as a final field?
```suggestion
        private static final DIVIDER = "--------------------------------------------------"
        // Code
        ...
        System.out.println(DIVIDER);
```

src/main/java/UI/UI.java
```diff
@@ -0,0 +1,68 @@
...
+    public void showErrorMessage(String message) {
+        System.out.println(message);
+    }
+
+    /**
+     * Displays a separator line followed by a newline to the console.
+     */
+    public void showNewLineSeparator() {
+        System.out.println("--------------------------------------------------\n");
+    }
```
valhrd: Can we try using a static final field?
```suggestion
    public void showNewLineSeparator() {
        System.out.println(DIVIDER + "\n");
    }
```

valhrd (commented):
Overall really well written code!

src/main/java/UI/Parser.java
keerthigkaarthik: This would throw an invalid index error if there was no " /by " but I could just wrap it with a try block. Your suggestions looks a lot more concise thanks!

src/main/java/Task/Task.java
```diff
@@ -0,0 +1,82 @@
...
+ * Abstract class representing a task with a completion status and formatting options.
+ * Hello!
+ */
+public abstract class Task {
+    protected static final DateTimeFormatter toSelfFormatter = DateTimeFormatter.ofPattern("MM-dd-yyyy HHmm");
+    protected static final DateTimeFormatter toUserFormatter = DateTimeFormatter.ofPattern("MMM dd yyyy HH:mm");
+
+    private final String task;
+    public String type;
+    private boolean completed;
```
zaidansani: since this is a boolean variable, it would be better to name it with prefixes - in this case, perhaps `isCompleted`

src/main/java/Task/Deadline.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    /**
+     * Returns a string representation of the deadline task formatted for saving to a file.
+     * The format includes the task type, completion status, task description, and deadline, separated by " | ".
+     * This string is used for persisting the task data in a structured text file, allowing for easy reloading of tasks.
+     *
+     * @return The formatted string for saving the deadline task to file.
+     */
+    @Override
```
zaidansani: there's a bit of repetition in the javadocs, perhaps to make it easier for you, if there are more tasks to be added, would be to utilise the `@inheritDocs` comment, and just add the specific difference in the overridden method after instead?

src/main/java/Task/TaskList.java
```diff
@@ -0,0 +1,124 @@
...
+        }
+        System.out.println();
+    }
+
+    /**
+     * Finds and returns a new TaskList containing all tasks that include the specified term.
+     * @param term The term to search for within task descriptions.
+     * @return A TaskList of all tasks containing the term.
+     */
+    public TaskList tasksContainingTerm(String term) {
```
zaidansani: generally, since this is a method, it should be a verb - `getTasksContainingTerm` to indicate that it is a method instead of a variable

src/main/java/Task/TaskList.java
```diff
@@ -0,0 +1,124 @@
...
+        }
+    }
+
+    /**
+     * Marks a task as not done.
+     * @param index The index of the task to mark as not done.
+     */
+    public void unmarkTask(int index) {
+        if (index >= 0 && index < tasks.size()) {
+            this.tasks.get(index).markAsUndone();
```
zaidansani: we can leave out the `this.` here to minimise noise, since there isn't another local `tasks` variable to confuse it with!

src/main/java/Bob/Bob.java
```diff
@@ -0,0 +1,41 @@
+package Bob;
+
+import Storage.Storage;
+import UI.UI;
+import UI.Parser;
+import UI.BotException;
+
+/**
```
zaidansani: your comments are excellent and descriptive in general! well done!

src/main/java/UI/Parser.java
```diff
@@ -0,0 +1,142 @@
+package UI;
```
zaidansani: package names should generally be lowercase

src/main/java/UI/UI.java
```diff
@@ -0,0 +1,71 @@
...
+
+import Task.TaskList;
+
+import java.util.Scanner;
+
+/**
+ * Handles user interactions such as displaying messages and receiving input.
+ * Hello!
+ */
+public class UI {
```
zaidansani: this should probably be `Ui` instead, in accordance to convention

src/test/java/UI/ParserTest.java
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Test class for Parser's handleDelete method to ensure it functions correctly
+ * with valid inputs, and throws the correct exceptions with invalid inputs.
+ */
+public class ParserTest {
+
+    /**
+     * Sets up the environment before each test, initializing the TaskList with predefined tasks.
+     */
+    @BeforeEach
```
zaidansani: i like this approach to the tests!

src/main/java/Task/ToDo.java
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Loads a ToDo task from a formatted string array obtained from splicing the save file
+     * during Storage's load method.
+     *
+     * @param properties The string array containing the task properties.
+     */
+    public static void load(String[] properties) {
+        ToDo newTodo = new ToDo(properties[2]);
+        if (Objects.equals(properties[1], "1")) {
```
zaidansani: out of curiosity, why use the objects.equals methods, and not just a `"1".equals(properties[1])`

src/main/java/Task/Task.java
zaidansani: just noted your method is name `isCompleted`, perhaps that can be replaced with `checkIfCompleted`?

src/main/java/Task/Task.java
```diff
@@ -0,0 +1,82 @@
...
+    public Task(String description) {
+        this.task = description;
+        this.completed = false;
+        TaskList.mainTaskList.addTask(this);
+    }
+
+    /**
+     * Marks the task as completed.
+     */
+    protected void markAsDone() {
```
zaidansani: since this is effectively a boolean method, it might be better to name this `setIsComplete` instead, based on coding convention!

src/main/java/Task/Deadline.java
zaidansani: specifically referring to this method since you have long comments for the abstract method and the overridden one as well, with repeated information

zaidansani (commented):
hi keerthig, excellent work so far - i think some areas of improvement could be your naming (package names and class name conventions).

your comments are EXCELLENT - they're comprehensive, and informative!

src/main/java/Storage/Storage.java
```diff
@@ -0,0 +1,99 @@
+package Storage;
```
skyl3-r: Package names should be in lower case in order to avoid conflict with classes or interfaces.

src/main/java/Bob/Bob.java
```diff
@@ -0,0 +1,41 @@
+package Bob;
```
skyl3-r: It might make more sense to place the storage, task and ui packages as well as other bob related files in the bob package.

src/main/java/UI/UI.java
```diff
@@ -0,0 +1,71 @@
...
+    static void taskAddedMsg() {
+        System.out.println("Sure! I've added that in for you.");
+        TaskList.mainTaskList.printList();
+    }
+
+    /**
+     * Displays a separator line to the console.
+     */
+    public void showSeparator() {
+        private static final DIVIDER = "--------------------------------------------------"
```
skyl3-r: Just curious, why is the constant DIVIDER defined in a non static method? What happens if showNewLineSeparator is run before showSeparator?

src/main/java/UI/Parser.java
```diff
@@ -0,0 +1,142 @@
...
+            System.out.printf("There is 1 task containing the word \"%s\".\n", arguments);
+        } else {
+            System.out.printf("There are %s tasks containing the word \"%s\".\n", searchResultList.getNumTasks(), arguments);
+        }
+    }
+
+    private static void handleMarkUnmark(String command, String arguments) throws BotException {
+        if (arguments.isEmpty()) throw new BotException("Please provide a task number.");
+        int taskIndex = Integer.parseInt(arguments) - 1;
+        if (taskIndex < 0 || taskIndex >= TaskList.mainTaskList.getNumTasks()) throw new BotException("That task does not exist!");
```
skyl3-r: I think the if statement should be formatted as
```suggestion
if (taskIndex < 0 || taskIndex >= TaskList.mainTaskList.getNumTasks()) {
    throw new BotException("That task does not exist!");
}
```
so that the line isn't too long and it is easier to read.

src/main/java/Storage/Storage.java
```diff
@@ -0,0 +1,99 @@
...
+
+import Task.Deadline;
+import Task.Event;
+import Task.TaskList;
+import Task.ToDo;
```
skyl3-r: Good job on listing the imported classes explicitly rather than using *.

src/main/java/Storage/Storage.java
```diff
@@ -0,0 +1,99 @@
...
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.Scanner;
+import java.io.FileNotFoundException;
+
+/**
+ * The Storage class handles loading and saving data to a specified file path.
+ * It supports operations to clear, save, and load data.
+ * Hello!!
```
skyl3-r: Good use of javadocs overall! Just not sure if the Hellos need to be there haha.

skyl3-r (commented):
Overall, I think you almost always stuck to style convention.  Good job!

PR #225, opened 2024-08-26
[josh1248] iP

josh1248:
# Grok
> Grok' means to understand so thoroughly that the observer becomes a part of the process being observed - to merge, to blend, to intermarry, to lose personal identity in group experience. - Robert A. Heinlein

Understand and [be one with productivity](https://github.com/nus-cs2103-AY2425S1/ip/pull/225) - with the Grok chatbot. Makes productivity so easy that you embody productivity. 🥇
- 0% hassle
- Minimalist. ~~No extra junk.~~
- 100% smooth

**Features**
1. Ability to add and remove 3 types of todos - tasks, deadlines, and events
2. Seamless text-based storage that is persistent
3. Ability to mark and unmark todos as required

`Grok` just checks all the boxes:
- [x] Lightweight
- [x] Amazing
- [x] `Grok`

```Java
public class Grok {
    public static void launchApp() {
        You.setProductivity(110);
    }
}
```


src/test/java/parser/ParserTest.java
```diff
@@ -0,0 +1,46 @@
+package parser;
+
+import commands.ExitCommand;
+import exceptions.GrokInvalidUserInputException;
+import org.junit.jupiter.api.BeforeAll;
```
zhou-colla, line 5: Would it looks neater if we separate the import statements for the import statements? 🤔

src/main/java/commands/Command.java
```diff
@@ -0,0 +1,39 @@
...
+    // The breaking of the abstraction barrier here is unavoidable as the command must terminate the entire program.
+    public boolean isExit() {
+        return isExitCommand;
+    }
+
+    /**
+     * Carry out the required actions of the command, including UI input, task list updating, and storage.
+     * Unifying under a single method allows for abstraction of the effects of the command from the caller.
+     */
+    public abstract void execute(TaskList tasks, Ui ui, Storage storage);
```
starrdye: Would it be better if we explains the functionalities of the parameters?

src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,152 @@
+package parser;
+
+import commands.*;
```
starrdye: Is the wildcard necessary here?

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,92 @@
+package storage;
+
+import exceptions.GrokInvalidUserInputException;
+import taskList.TaskList;
+import tasks.Deadline;
+import tasks.Event;
+import tasks.Task;
```
starrdye: I like how you sorted the import statements in order.

src/main/java/parser/Parser.java
josh1248: thank you for pointing this out - will resolve it!

src/main/java/tasks/Deadline.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    /**
+     * Initializes a deadline with the following inputs:
+     * @param description title, or name, of deadline
+     * @param due deadline of date which MUST be in ISO yyyy-mm-dd format
+     * @param isDone whether the deadline has been met or not
+     * @throws GrokInvalidUserInputException if user input is invalid - empty description, or non-ISO dates
+     */
+    public Deadline(String description, String due, Boolean isDone) throws GrokInvalidUserInputException {
```
yastsc, line 38: Might be better to use the primitive type boolean instead of the reference type Boolean as argument type.

src/main/java/ui/Ui.java
yastsc, line 1: All the method names in this class are very well named, and accurately describe what the method does, so good job!

src/main/java/tasks/Task.java
```diff
@@ -0,0 +1,66 @@
...
+            throw new GrokInvalidUserInputException("Grok.Task description cannot be empty!");
+        }
+        this.description = description;
+        this.isDone = isDone;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); // mark done task with X
+    }
+
```
yastsc, line 43: Perhaps you can write more JavaDoc comments for the public methods?

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,101 @@
...
+
+    public Storage(String filePath) {
+        this.filePath = filePath;
+    }
+
+    /**
+     * Converts a compliant text string from a text file into a list of tasks.
+     * @return an ArrayList of tasks.
+     */
+    public ArrayList<Task> parseTextStorage() {
```
yastsc: loadTextStorage() may be a better name as the method doesn't just parse but also loads the stored tasks into an ArrayList that is returned.

yastsc (commented):
Overall, your naming of classes, methods and variables are really good! There seem to be only one or two method names that can be improved on. The coding standard is also mostly followed, so well done :)

PR #227, opened 2024-08-26
[The0nlyJuan] iP

The0nlyJuan:
### Krona Chatbot

> "Your tasks are safe with Krona. Never forget another task!" – Krona Team

Krona helps you manage tasks efficiently and effectively. It's designed to make tracking your to-do list a breeze.

- Easy to use

- Fast and efficient

- Customizable task management

All you need to do is:

1. Download Krona [here](https://github.com/The0nlyJuan/ip/tree/v0.1).
2. Run the program by double-clicking.
3. Start adding your tasks.
4. Let Krona manage them for you! :wink:

And it is **FREE**!

Features:

- [x] Managing tasks
- [x] Setting deadlines 
- [ ] Task reminders (coming soon)

```java
public class Main {
    }
}

### CODE REVIEW

File: src/main/java/Main.java [hunk 1]
```diff
@@ -0,0 +1,34 @@
...
+    private Krona krona = new Krona("data/tasks.txt");
+
+    @Override
+    public void start(Stage stage) {
+        try {
+            FXMLLoader fxmlLoader = new FXMLLoader(Main.class.getResource("/view/MainWindow.fxml"));
+            AnchorPane ap = fxmlLoader.load();
+            Scene scene = new Scene(ap);
+            stage.setScene(scene);
+            stage.setMinHeight(220);
```
josh1248, line 25: I believe this could be represented as a constant in the Main class, for example private static int MIN_HEIGHT_PX = 220. This would avoid the issue of magic constants, and explains the units involved in setting the height as well.

src/main/java/krona/command/AddCommand.java
```diff
@@ -0,0 +1,44 @@
...
+     * @param tasks The task list that the command operates on.
+     * @param ui The UI component that handles interactions with the user.
+     * @param storage The storage component that handles saving the updated task list.
+     * @throws KronaException If an error occurs while saving the task list.
+     */
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws KronaException {
+        tasks.addTask(task);
+        StringBuilder output = new StringBuilder();
+        output.append("Got it. I've added this task:\n");
```
josh1248, line 37: Would this belong in the purview of the UI as it is feedback to the user? Could consider.

src/main/java/krona/command/DeleteCommand.java
```diff
@@ -0,0 +1,52 @@
...
+        try {
+            Task task = tasks.get(taskIndex);
+            tasks.deleteTask(taskIndex);
+
+            StringBuilder output = new StringBuilder();
+            output.append("Noted. I've removed this task:\n");
+            output.append(task.toString()).append("\n");
+            output.append("Now you have ").append(tasks.size()).append(" tasks in the list.");
+
+            storage.save(tasks);
```
josh1248, line 44: Love the use of line breaks here to separate the different 'phases' of the execute.

src/main/java/krona/parser/Parser.java
```diff
@@ -0,0 +1,131 @@
+package krona.parser;
+
+import krona.command.AddCommand;
```
josh1248, line 3: nice alphabetical order to easily identify the import.

src/main/java/krona/parser/Parser.java
```diff
@@ -0,0 +1,131 @@
...
+public class Parser {
+
+    /**
+     * Parses the user's input command and returns the corresponding Command object.
+     *
+     * @param input The user input as a string.
+     * @return The Command object corresponding to the user input.
+     */
+    public static Command parse(String input) {
+        String[] words = input.split(" ", 2);
```
josh1248, line 32: this is not coding standard related, but i think this may detract from the project scope if the task is, for example, `todo borrow book` (example given from week 2 ip project requirements).

src/main/java/krona/task/Event.java
```diff
@@ -0,0 +1,49 @@
...
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+/**
+ * Represents an Event with a timeline from a start date
+ * and time to an end date and time.
+ */
+public class Event extends Task {
+    private static final DateTimeFormatter formatter = DateTimeFormatter.ofPattern("d/M/yyyy HHmm");
+    protected LocalDateTime startDateTime;
```
josh1248, line 13: since you have getter methods, it would be appropriate to place the access attributes here as `private`.

src/main/java/krona/storage/Storage.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.ArrayList;
+import java.util.Scanner;
+
+import krona.exception.KronaException;
+import krona.parser.Parser;
+import krona.task.Task;
+import krona.task.TaskList;
+
+
+/**
```
josh1248, line 15: Very helpful class and method explanations here.

src/main/java/krona/ui/Ui.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public Ui(Stage stage) {
+        this.stage = stage;
+    }
+
+    public void goodbye() {
+        setCombinedMessage("Bye. Hope to see you again soon!");
+    }
+
+    public void exitApp() {
```
yimjianbing, line 36: Really like this implementation to exit from the javafx application after the user input bye! Good job!

src/main/resources/view/DialogBox.fxml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<?import javafx.geometry.Insets?>
+<?import javafx.scene.control.Label?>
+<?import javafx.scene.image.ImageView?>
+<?import javafx.scene.layout.HBox?>
+
+<fx:root alignment="TOP_RIGHT" maxHeight="1.7976931348623157E308" maxWidth="1.7976931348623157E308" prefWidth="400.0" stylesheets="@../css/dialog-box.css" type="javafx.scene.layout.HBox" xmlns="http://javafx.com/javafx/22" xmlns:fx="http://javafx.com/fxml/1">
```
yimjianbing: Small nitpick but I don't think your .fxml is working? Your .fxml is stating it using java SDK vers 22, usually because every time you use scenebuilder to save the edits onto your .fxml file, it will set the version as xmlns="http://javafx.com/javafx/22"  instead of xmlns="http://javafx.com/javafx/17", so need to edit this file every time you use scenebuilder

src/main/resources/view/MainWindow.fxml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<?import javafx.scene.control.Button?>
+<?import javafx.scene.control.ScrollPane?>
+<?import javafx.scene.control.TextField?>
+<?import javafx.scene.layout.AnchorPane?>
+<?import javafx.scene.layout.VBox?>
+
+<AnchorPane maxHeight="-Infinity" maxWidth="-Infinity" minHeight="-Infinity" minWidth="-Infinity" prefHeight="600.0" prefWidth="400.0" stylesheets="@../css/main.css" xmlns="http://javafx.com/javafx/22" xmlns:fx="http://javafx.com/fxml/1" fx:controller="MainWindow">
```
yimjianbing: same here with the java version being 22 instead of 17!

src/test/java/krona/KronaTest.java
```diff
@@ -0,0 +1,4 @@
+package krona;
+
+public class KronaTest {
```
yimjianbing, line 3: I think having a JUnit test for the Krono app itself might be a bit impractical? Since you would have to isolate many of its dependent classes with SUTs and test drivers, but definitely not impossible

src/main/java/krona/task/Deadline.java
```diff
@@ -0,0 +1,40 @@
...
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+/**
+ * Represents a task with a deadline that the user
+ * can input date and time.
+ */
+public class Deadline extends Task {
+    private static final DateTimeFormatter formatter = DateTimeFormatter.ofPattern("d/M/yyyy HHmm");
```
yimjianbing: Maybe you can change the formatter name to be in all caps instead, since it has the final modifier, it means its a constant, so we have to put its name in all caps according to the coding standard

src/main/java/krona/task/Event.java
```diff
@@ -0,0 +1,49 @@
...
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+/**
+ * Represents an Event with a timeline from a start date
+ * and time to an end date and time.
+ */
+public class Event extends Task {
+    private static final DateTimeFormatter formatter = DateTimeFormatter.ofPattern("d/M/yyyy HHmm");
```
yimjianbing: Since this variable, formatter, is also final and thus a constant, maybe you should change its name to FORMATTER as well

src/main/java/krona/command/MarkCommand.java
```diff
@@ -0,0 +1,49 @@
...
+     * @param storage The storage component that handles saving the updated task list.
+     * @throws KronaException If the task index is out of bounds or an error occurs while saving the task list.
+     */
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws KronaException {
+        try {
+            Task task = tasks.get(taskIndex);
+            task.markDone();
+
+            StringBuilder output = new StringBuilder();
```
yimjianbing, line 39: I really like the usage of StringBuilders here! Good job!

src/main/java/Main.java
```diff
@@ -0,0 +1,34 @@
...
+import javafx.scene.layout.AnchorPane;
+import javafx.stage.Stage;
+import krona.Krona;
+
+/**
+ * A GUI for Duke using FXML.
+ */
+public class Main extends Application {
+
+    private Krona krona = new Krona("data/tasks.txt");
```
yimjianbing, line 15: I'm not sure if its my side, but when I used the Bye Command with your Krono, it was unable to save the tasks into a task.txt, and I couldn't find a tasks.txt in the directory, so maybe you can add it to your git? Unless I'm blind in which I sincerely apologise!

yimjianbing (commented):
Overall the code is super clean and readable. I could easily navigate around the different classes, and the methods/variable names are all really well done. I knew what everything did, and I really enjoy how some of the features were implemented! Great job!

src/main/java/Main.java
The0nlyJuan: Hey! Am i supposed to add the data/tasks.txt to git? i just checked and it was under ignored files because i assume its supposed to be for each individual computer

src/main/resources/view/DialogBox.fxml
The0nlyJuan: Oh okay I did not even know this. Thank you!

PR #231, opened 2024-08-26
[S-K-Y-Light] iP

S-K-Y-Light:
# Sora

> They Can Take Your World, They Can Take Your Heart, Cut You Loose From All You Know. But If It's Your Fate, Then Every Step Forward Will Always Be A Step Closer To Home – Sora ([Kingdom Hearts 3](https://www.kingdomhearts.com/3/us/home/))

Sora stresses your mind by reminding things you need to do. It's,
- text-based (Unfriendly to GUI Users)
- easy to learn (After One Month of Usage)
- ~~FAST~~ Negligibly Fast to use (Runs in O(N!) Runtime)

All you need to do is,
1. download it from [here](https://github.com/S-K-Y-Light/ip).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:
- [x] Managing Tasks
- [ ] Graphical User Interface (coming soon)
- [ ] Automatically Doing your Tasks (coming never)

---

If you are a Java Programmer, you can use it to practice your Java too. Here's a main method that throws a SoraException:
```
public class Main {
        throw new SoraException();
    }
}
```


src/main/java/sora/Parser.java
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * Returns an ArrayList that parses the User's command.
+     * Identifies mainCommand and additional details in the User's input.
+     *
+     * @param command User's command / input.
+     * @return ArrayList containing parsed command.
+     */
+    public static ArrayList<String> parse(String command) {
+        ArrayList<String> parsedResult = new ArrayList<>();
+        String[] parse_1 = command.split(" ", 2);
```
Nicholaskoh1: Perhaps this could be better named? Maybe something like parseTemp or parsedArr instead of using _

src/main/java/sora/Storage.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    /**
+     * Returns String of all Task Details divided by " | ".
+     * It is a Helper Function for saveTaskList(TaskList)
+     *
+     * @param taskList TaskList.
+     * @return String of all Task Details divided by " | ".
+     */
+    private String saveTaskString(TaskList taskList) {
```
Nicholaskoh1: Perhaps this could be better named? I initially interpreted this as 'saving task strings' and not 'getting the task strings to be saved'.

src/main/java/sora/Tasks/Deadline.java
```diff
@@ -0,0 +1,55 @@
...
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.List;
+
+/**
+ * Deadline is a Task with a deadline.
+ */
+public class Deadline extends Task {
+    /** Deadline of the deadline in String Format */
+    protected String by;
```
Nicholaskoh1, line 12: Perhaps this could be better named? 'by' seems a bit ambigious on what it stands for.

src/main/java/sora/Tasks/Task.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * Returns True if this description contains the substring.
+     * This description and substring's casing are ignored.
+     *
+     * @param s User's Substring.
+     * @return Boolean value.
+     */
+    public boolean descriptionContainsString(String s) {
```
Nicholaskoh1, line 61: Perhaps could be better named to suit a boolean method?

Nicholaskoh1 (commented):
Just some nitpicks on the names of some of the methods and variables.

src/main/java/sora/Parser.java
S-K-Y-Light: Nice Catch. Will Rename.

src/main/java/sora/Storage.java
S-K-Y-Light: Sounds solid. Will Rename.

src/main/java/sora/Parser.java
```diff
@@ -0,0 +1,47 @@
...
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.util.ArrayList;
+import java.util.Arrays;
+
+/**
+ * Parser parses the User's input.
+ * It deals with making sense of the user command.
+ * It identifies mainCommand and additional details in the User's input.
+ */
```
YANYAN-w, line 13: I like that your JavaDocs is very comprehensive!

src/main/java/sora/Storage.java
```diff
@@ -0,0 +1,131 @@
...
+        this.filePath = filePath;
+    }
+
+    /**
+     * Returns String of all Task Details divided by " | ".
+     * It is a Helper Function for saveTaskList(TaskList)
+     *
+     * @param taskList TaskList.
+     * @return String of all Task Details divided by " | ".
+     */
```
YANYAN-w: I like that the overhead comments for public methods are comprehensive!

src/main/java/sora/Storage.java
```diff
@@ -0,0 +1,131 @@
...
+     * @param fileTaskDetails Formatted task details in this Storage instance's filePath.
+     * @param taskList TaskList.
+     * @throws SoraException If line in this Storage instance's filePath is wrongly formatted.
+     */
+    private void loadTaskListHelper(int lineNumber, String fileTaskDetails, TaskList taskList) throws SoraException {
+        String[] parsedFileTaskDetails = fileTaskDetails.split(" \\| ", 5);
+        switch (parsedFileTaskDetails[0]) {
+            case "T":
+                if (parsedFileTaskDetails.length != 3) {
+                    throw new SoraException("\tSora is unable to read file from line " + lineNumber);
```
YANYAN-w: Perhaps it is better to follow the coding standard that there is no indentation for case clauses 🤔

src/main/java/sora/Tasks/Task.java
YANYAN-w, line 61: Perhaps the method name can be changed to sound more like a boolean method?

src/main/java/sora/Storage.java
S-K-Y-Light: Good Catch. I wonder how I missed that 🤔 
Will Fix 👍

PR #232, opened 2024-08-26
[wang-h-z] iP

wang-h-z:
Can't seem to remember your schedule? Friday's got your back. Friday is: 
* easy to use
* ~easy~ SUPER EASY to learn
* **FAST** :fire:

How to use Friday: 
1. clone the repo from [here](https://github.com/wang-h-z/ip)
2. follow the README.md file 
3. enjoy!

Best part about Friday? Its completely **FREE!**

Features: 
- [x] Managing tasks
- [ ] Filtering tasks (coming soon)
- [ ] Finding tasks (coming soon) 

Here's a short snippet of Friday's codebase! 

```java
        Friday bot = new Friday("data/ui.Friday.txt");
        bot.start();
    }
```


src/main/java/common/BotMessage.java
```diff
@@ -0,0 +1,15 @@
...
+    private final String spacing = "____________________________________________________________";
+    private final String description;
+
+    public BotMessage(String description) {
+        this.description = description;
+    }
+
+    @Override
+    public String toString() {
+        return '\t' + this.spacing + '\n' + this.description + '\n' + '\t' + this.spacing + "\n";
```
Dinoman44: Small nit, you could have just done `... this.description + '\n\t' + ...` but eh.

src/main/java/exceptions/FridayException.java
```diff
@@ -0,0 +1,8 @@
+package exceptions;
```
Dinoman44, line 1: Seems like all these exceptions are related to Task handling. Why don't you create a [sub-package](https://www.refreshjava.com/java/sub-packages-in-java) inside of the `tasks` package so you can then import the exceptions as `import tasks.exceptions.<exception>`?

src/main/java/tasks/Todo.java
```diff
@@ -0,0 +1,29 @@
...
+        return "[T]" + super.toString();
+    }
+
+    /**
+     * Returns a Todo Task given a String that represents a Todo Task.
+     *
+     * @param str The String that represents the Todo Task.
+     * @return
+     */
+    public static Todo fromString(String str) {
```
Dinoman44, line 33: Small nit, you seem to be missing the `@Override` annotation here.

src/main/java/tasks/Event.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public Event(String description, String from, String to) {
+        super(description);
+        this.from = from;
+        this.to = to;
+        try {
+            this.fromDate = DateTimeParser.parse(from);
+        } catch (IllegalArgumentException e) {
+            //No correct format, deadline only recorded as a String
+            this.fromDate = null;
```
Dinoman44, line 41: Would it be a better idea to reprompt the user for input instead of setting the `fromDate`/`toDate` as `null`? This way working with DateTimes would be easier and you can allow for more functionality (like searching for tasks on a day).

src/main/java/tasks/Event.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    /**
+     * Returns an Event given a String that represents the Event task.
+     *
+     * @param str The String that represents the Event Task.
+     * @return
+     */
+    public static Event fromString(String str) {
+        boolean isDone = str.contains("[X]");
```
Dinoman44, line 80: What if some weird case inputs `"[X]"` as the task description? It might be better to confirm that the `"[X]"` is at the right place (i.e. from index 3 to 5) if at all present.

src/main/java/tasks/Task.java
```diff
@@ -0,0 +1,51 @@
...
+        } else if (str.startsWith("[D]")) {
+            return Deadline.fromString(str);
+        } else if (str.startsWith("[E]")) {
+            return Event.fromString(str);
+        } else {
+            throw new IllegalArgumentException("Unknown task type: " + str);
+        }
+    }
+
+    public String getDescription() {
```
Dinoman44, line 123: Small nit: again missing `@Override` annotation.

src/main/java/common/DateTimeParser.java
```diff
@@ -0,0 +1,53 @@
+package common;
```
Dinoman44, line 1: Since this is a parser, why not have it under the `parser` package instead?

src/main/java/common/TaskList.java
```diff
@@ -0,0 +1,44 @@
+package common;
```
Dinoman44, line 1: Since this is related to tasks, why not have it in the `tasks` package?

src/main/java/ui/Ui.java
```diff
@@ -0,0 +1,91 @@
...
+    public void showOutput(String str) {
+        BotMessage output = new BotMessage(str);
+        System.out.println(output);
+    }
+
+    public String readCommand() {
+        return this.scanner.nextLine();
+    }
+
+    public void listOutput(TaskList list) {
```
Dinoman44, line 91: Naming note: It's better to have function names as verbs, ex: `outputList` or `outputTaskAddition`.

src/main/java/commands/ByeCommand.java
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * The user interface will show an ending message when this method is called.
+     *
+     * @param list List which stores all Tasks in the chatbot.
+     * @param ui The interface which the user will be interacting with.
+     * @param storage Stores previous and current Task objects.
+     */
+    @Override
+    public void execute(TaskList list, Ui ui, Storage storage) {
+        ui.showEnd();
```
Dinoman44, line 23: I may have missed it, but where is the part where the tasks get stored from the tasklist to the text file?

src/main/java/commands/DateCommand.java
```diff
@@ -0,0 +1,76 @@
...
+                Event e = (Event) t;
+                if (e.getDate() != null && e.getDate().isBefore(userTime)) {
+                    dateList.add(e);
+                }
+            }
+        }
+        ui.dateOutput(userTime.format(formatter), list);
+    }
+
+    /**
```
Dinoman44: For overridden functions I think you can just have the JavaDoc comment in the top-level class, then no need to re-define it for the classes that inherit from that class.

src/main/java/commands/ByeCommand.java
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    /**
+     * The ByeCommand signifies that the user would like to stop the chatbot.
+     *
+     * @return false
```
Dinoman44, line 29:
```suggestion
     * @return true
```

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,71 @@
...
+     * Writes the Tasks contained in the TaskList to a new .txt file.
+     * <p>
+     * Throws an IOException is writing to the .txt file fails.
+     *
+     * @param list A TaskList containing all tasks.
+     */
+    public void saveTasks(TaskList list) {
+        try (PrintWriter writer = new PrintWriter(new FileWriter(FILE_PATH))) {
+            for (Task task : list.getList()) {
+                writer.println(task.toString());
```
Dinoman44, line 67: I did not know you could do that, I might try this out in my code. I was using the `java.io.FileWriter` class and the `FileWriter::write` method.

data/Friday.txt
```diff
@@ -0,0 +1,7 @@
+[D][ ] Task (by: Dec 02 2019, 6:00 pm)
+[E][ ] Task (from: Dec 02 2019, 6:00 pm to: Dec 02 2019, 8:00 pm)
+[D][ ] Task (by: Dec 02 2019, 12:00 am)
+[T][ ] Task
+[D][ ] Task (by: 23-02-2020 1800)
+[D][ ] Task (by: Feb 06 2002, 6:00 pm)
+[T][ ] wth
```
Dinoman44:
```suggestion
[T][X] wth
```

build.gradle
```diff
@@ -0,0 +1,41 @@
...
+        showStandardStreams = false
+    }
+}
+
+application {
+    mainClass.set("Friday")
+}
+
+shadowJar {
+    archiveBaseName = "duke"
```
Dinoman44:
```suggestion
    archiveBaseName = "friday"
```

build.gradle
```diff
@@ -0,0 +1,41 @@
...
+    archiveClassifier = null
+}
+
+run{
```
Dinoman44, line 59:
```suggestion
run {
```

Dinoman44 (commented):
Looks solid, some minor refactoring and coding convention changes needed. Good luck!

src/main/java/tasks/Deadline.java
```diff
@@ -0,0 +1,49 @@
+package tasks;
+
+import java.time.LocalDateTime;
+
+import common.DateTimeParser;
+
+public class Deadline extends Task {
+
+    private String by;
+    private LocalDateTime byDate;
```
ulricolo7, line 16: Would it be better if the variables here are named more descriptively, instead of 'by' and 'byDate'?

src/main/java/tasks/Event.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.time.LocalDateTime;
+
+import common.DateTimeParser;
+
+public class Event extends Task {
+    private String from;
+    private String to;
+    private LocalDateTime fromDate;
+    private LocalDateTime toDate;
```
ulricolo7, line 18: Similar comment for 'from', 'to', 'fromDate', and 'toDate'! Would it be better if it was named more descriptively?

src/main/java/ui/Friday.java
```diff
@@ -0,0 +1,62 @@
...
+import common.TaskList;
+
+import exceptions.FridayException;
+
+import parser.Parser;
+
+import storage.Storage;
+
+public class Friday {
+    private final BotMessage header = new BotMessage("""
```
ulricolo7, line 16: Perhaps add a descriptor as to what header this is for? The variable 'ending' too

ulricolo7 (commented):
LGTM! Just pointing out some naming conventions! Generally could be a bit more descriptive haha

PR #237, opened 2024-08-26
[kaajinn] iP

kaajinn:
# Applemazer
> "His micro adjustments are crazy!" – Anonymous

Applemazer frees your mind of having to remember things you need to do. It's,
- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,
1. download it from [here](https://github.com/kaajinn/ip).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)
#

```Java
public class Applemazer {
        new Applemazer("./data/Applemazer.ser").process();
    }
}
```


verakohh:
### Issue 1: Invalid Date Handling for Events

I noticed that the current implementation allows the creation of events with invalid dates, such as when the 'to' date is earlier than the 'from' date. For example:

``` 
event meeting /from 2022-12-12 /to 2021-12-12
Got it. I've added this task: 
    [E][ ] meeting (from: 12 Dec 2022 to: 12 Dec 2021) 
Now you have 5 tasks in the list. 

Saving data...
Save successful.
```

This could lead to confusion and potentially incorrect data being stored. I suggest adding validation to check that the 'to' date is not earlier than the 'from' date before allowing the task to be created.

### Issue 2: Missing Hi/Hello Command

It looks like the application might be missing a response for a `hi` or `hello` command. Since users might naturally start an interaction with a greeting, adding this command could improve user experience. For instance, a simple response like "Hi there! How can I assist you today?" would be great.

---

### Additional Feedback:
- I really like how the time component is optional for deadlines and events!

src/main/java/applemazer/Parser.java
```diff
@@ -0,0 +1,177 @@
...
+     * @return Command object that is ready to execute the user command.
+     * @throws Exception Thrown when parsing fails due to incorrect user input format.
+     */
+    public Command parse(String command) throws Exception {
+        String desc;
+        String[] split;
+        int taskNumber;
+
+        return switch (command) {
+        case "bye" -> {
```
verakohh, line 49: Missing Hi/Hello Command

It looks like the application might be missing a response for a `hi` or `hello` command. Since users might naturally start an interaction with a greeting, adding this command could improve user experience. For instance, a simple response like "Hi there! How can I assist you today?" would be great.

src/main/java/commands/EventCommand.java
```diff
@@ -0,0 +1,62 @@
...
+     * @param storage The storage object containing the filepath which the chatbot saves to and loads from.
+     */
+    @Override
+    public void execute(TaskList tasks, Storage storage) {
+        try {
+            Task task = new Event(desc, from, to);
+            tasks.add(task);
+            task.printTaskAddedMessage(tasks.size());
+            storage.save();
+        } catch (DateTimeException e) {
```
verakohh, line 51: I noticed that the current implementation allows the creation of events with invalid dates, such as when the 'to' date is earlier than the 'from' date. For example:

``` 
event meeting /from 2022-12-12 /to 2021-12-12
Got it. I've added this task: 
    [E][ ] meeting (from: 12 Dec 2022 to: 12 Dec 2021) 
Now you have 5 tasks in the list. 

Saving data...
Save successful.
```

This could lead to confusion and potentially incorrect data being stored. I suggest adding validation to check that the 'to' date is not earlier than the 'from' date before allowing the task to be created.

verakohh (commented):
I really like how the time component is optional for deadlines and events!

src/main/java/applemazer/Applemazer.java
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Main class that runs the Applemazer chatbot.
+ */
+public class Applemazer {
+    protected static TaskList tasks;
+    protected static Scanner sc = new java.util.Scanner(System.in); // For text-based UI.
+    private static Storage storage;
+    private final Parser parser; // For text-based UI.
+    private final Ui ui;
+    private boolean isProcessing = true;
```
ashea10, line 20: Good job on being consistent with the naming standard for booleans, noticed these elsewhere as well!

src/main/java/applemazer/Applemazer.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    public Ui getUi() {
+        return this.ui;
+    }
+
+    public boolean shouldExit() {
+        return !this.isProcessing;
+    }
+
```
ashea10: Missing JavaDoc comments, would be great if you could add it in for all methods/classes.

src/main/java/applemazer/Parser.java
```diff
@@ -0,0 +1,177 @@
...
+        }
+        };
+    }
+
+    private void parseByeCommand() throws Exception {
+        String desc = sc.nextLine().trim();
+        if (!desc.isEmpty()) {
+            throw new Exception("OOPS!!! The description of bye should be empty. ");
+        }
+    }
```
ashea10, line 114: Would have been great if you have made use of custom exception(s).

src/main/java/applemazer/Storage.java
```diff
@@ -0,0 +1,94 @@
...
+            System.err.println("Save unsuccessful. File might be corrupted.\n");
+        }
+    }
+
+    /**
+     * Loads task list information from the stored filepath.
+     * <p>
+     * If the stored filepath does not exist, a new task list will be returned.
+     */
+    @SuppressWarnings("unchecked")
```
ashea10, line 73: To avoid the use of `SuppressWarnings` here, maybe you could use a method to convert the format read from the file to the format of a "task", so that we can do something like `Task.toProperFormat(str.readObject)`.

src/main/java/applemazer/Ui.java
```diff
@@ -0,0 +1,85 @@
...
+import tasks.Task;
+
+/**
+ * Class that handles chatbot messages.
+ */
+public class Ui {
+    /**
+     * Constructor for the Ui object.
+     */
+    public Ui() {}
```
ashea10: Empty constructor is present by default in Java. If it is the only constructor, it isn't required, though not wrong if it is there as well.

src/main/java/commands/Command.java
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * Returns {@code true} if the chatbot should continue running after executing the command.
+     * @return Either {@code true} or {@code false} depending on subclass implementation.
+     */
+    public abstract boolean continueProcessing();
+
+    /**
+     * Represents commands that handle integer input.
+     */
+    public enum IntegerCommands {
```
ashea10, line 34: Good job on using Enums!

src/main/java/tasks/Task.java
```diff
@@ -0,0 +1,56 @@
...
+     */
+    public void setUndone() {
+        isDone = false;
+    }
+
+    /**
+     * Returns the task description.
+     * @return The task description.
+     */
+    @Override
```
ashea10, line 48: Both methods are doing the same thing it seems, perhaps you can remove one and just stick with toString.

ashea10 (commented):
LGTM, good job! Just a few nits on the design choices.

PR #279, opened 2024-08-27
[reginateh] iP

reginateh:
# Otto :owl:
> Otto would rather be napping, but he supposes he can help you with your tasks.

Otto helps you remember things you need to do, although he doesn't want to. He is, 

- text-based
- easy to learn
- ~~FAST~~ *SUPER FAST* to use

All you need to do is,

2. double-click it.
3. add your tasks.
4. let it manage your tasks for you 😄 

## Features

 - [x] Managing tasks
 - [ ] Reminders (coming soon)

---
```Java
public class Main {
    }
}
```


src/main/java/Task.java
```diff
@@ -0,0 +1,19 @@
+abstract public class Task {
```
mattisongjj: I like the use of an abstract Task class that all other tasks inherit from, makes it easier in the case we want to add future tasks

mattisongjj: Perhaps this can be further abstracted by placing all relevant code in a 'task' package

src/main/java/Otto.java
```diff
@@ -0,0 +1,113 @@
...
+        this.displayMsg(OttoResponses.bye);
+    }
+
+    private void displayMsg(String msg) {
+        printLine();
+        System.out.println(msg.indent(4));
+        printLine();
+    }
+
+    private void displayTaskList() {
```
mattisongjj: Perhaps tasks related methods could be placed in the TaskList.java file rather than in this file

src/main/java/OttoResponses.java
```diff
@@ -0,0 +1,27 @@
+public final class OttoResponses {
```
mattisongjj: I like the use of a class that encapsulates the responses

mattisongjj: Could look into changing the names of the variables here to screaming snake case, since the otto responses are constant

mattisongjj (commented):
LGTM

src/main/java/Otto.java
```diff
@@ -0,0 +1,113 @@
...
+                case "delete":
+                    this.deleteTask(Parser.parseDeleteTask(command));
+                    break;
+                default:
+                    throw new OttoException(OttoResponses.unknown);
+            }
+        } catch (OttoException e) {
+            this.displayMsg(e.getMessage());
+        }
+    }
```
L-rrrr: I like how you follow the coding standard and the Egyptian style to make the code more readable

src/main/java/OttoResponses.java
```diff
@@ -0,0 +1,27 @@
...
+
+    public static final String complete = "Well, finally. You finished something.\n";
+    public static final String incomplete = "So you didn't finish that task. Try to get it done so Otto can rest easy.\n";
+    public static final String unknown = "Otto doesn't recognize this command. Speak English.";
+    public static final String indexError = "You need to give Otto an index that is in the list. Can't you even count?";
+    public static final String deleteError = "You need to tell Otto what task you want to delete. Not like Otto cares though.";
+    public static final String markError = "You need to tell Otto what task you want to mark. Not like Otto cares though.";
+    public static final String eventError = "The format for event is wrong. Missing description, start time or end time.";
+    public static final String deadlineError = "The format for deadline is wrong. Missing description or deadline.";
+    public static final String todoError = "Todo must contain a description.";
```
L-rrrr: I like that you make the strings final here so that they cannot be modified in other classes

src/main/java/Task.java
```diff
@@ -0,0 +1,19 @@
...
+    private String description;
+    private boolean isComplete;
```
L-rrrr: I like how you name your boolean field

src/main/java/Parser.java
```diff
@@ -0,0 +1,71 @@
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
```
L-rrrr: I like how you import and there is no use of wildcard

L-rrrr (commented):
Looks all good!
Perhaps can add some javadoc for different class and methods.

PR #283, opened 2024-08-27
[AbdulrahmanAlRammah] iP

AbdulrahmanAlRammah:
### Twilight 😁 

> "Every big change starts with a small step"

Twilight helps **_you_** keep track of:
- todo
- event
- deadlines

How to use:
1. launch program
2. enter commands as needed
3. say "bye" to close

Remaining improvements:
- [ ] Add help command
- [ ] Add GUI

For issues:[link](https://github.com/nus-cs2103-AY2425S1/forum)

Example of constructor for a command class:
```java
public AddCommand(String details) {
    this.details = details;
}
```


src/main/java/twilight/AddCommand.java
```diff
@@ -0,0 +1,36 @@
+package twilight;
+
+import java.io.IOException;
+
+/**
+ * Represents a Command to add a particular instance of an event.
+ */
+public class AddCommand extends Command {
```
yizhong187: Should `AddCommand` be renamed to `AddEventCommand` for **add**ed clarity?

src/main/java/twilight/Deadline.java
```diff
@@ -0,0 +1,45 @@
...
+        this.deadline = LocalDate.parse(deadline);
+    }
+
+    @Override
+    public String toString() {
+        return "[D]" + super.toString() + " by: " + this.deadline.format(DateTimeFormatter.ofPattern("MMM d yyyy"));
+    }
+
+    @Override
+    public String toStorage() {
```
yizhong187: Maybe `toStorageString` will be a more apt name as it is returning a string? Similarly for the other Command subclasses.

src/main/java/twilight/Event.java
```diff
@@ -0,0 +1,26 @@
+package twilight;
+
+public class Event extends Task {
+    protected String starttime;
```
yizhong187: Should `starttime` be in pascalCase like `startTime`?

src/main/java/twilight/MarkingCommand.java
```diff
@@ -0,0 +1,32 @@
+package twilight;
+
+import java.io.IOException;
+
+public class MarkingCommand extends Command {
+    protected int type;
```
yizhong187, line 9: Since type only has the value of 1 and 2, will it be better to declare an enum for it?

src/main/java/twilight/Parser.java
```diff
@@ -0,0 +1,53 @@
...
+    public static Command parse(String input) throws InvalidInputException {
+        try {
+            String[] split = input.split(" ", 2);
+            String command = split[0];
+            String details = split[1];
+            int i = 0;
+            while (!COMMANDS[i].equals(command)) {
+                i++;
+            }
+            if (i == 0) {
```
yizhong187, line 26: Would using a switch statement be cleaner here? Also, perhaps checking if the command equate to the exact string command (i.e. `command == "event"`) would make your code more readable?

src/main/java/twilight/AddCommand.java
```diff
@@ -0,0 +1,36 @@
...
+    protected String details;
+
+    public AddCommand(int type, String details) {
+        this.type = type;
+        this.details = details;
+    }
+
+    @Override
+    public void execute(TaskList tasks, Storage storage) throws InvalidInputException {
+        String addition = "";
```
yizhong187: Is it possible to give the string a more descriptive name?

src/main/java/twilight/AddCommand.java
```diff
@@ -0,0 +1,36 @@
...
+        if (type == 3) {
+            addition = tasks.add(new Todo(details));
+        } else if (type == 4) {
+            String[] split = details.split(" /from | /to ");
+            addition = tasks.add(new Event(split[0], split[1], split[2]));
+        } else {
+            String[] split = details.split(" /by ");
+            addition = tasks.add(new Deadline(split[0], split[1]));
+        }
+        System.out.println(addition);
```
yizhong187: Since you already created the `Ui` abstraction with a `print`, should you use that instead? Or should you abstract out all the UI handling into the `ui` object instantitated in the Twilight` class?

src/main/java/twilight/Parser.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.lang.IndexOutOfBoundsException;
+
+/**
+ * Represents a class used for transforming user input to a command to be carried out by chatbot.
+ */
+public class Parser {
+    private static final String[] COMMANDS = new String[]{"list", "mark", "unmark", "todo", "event", "deadline", "delete", "find", "bye"};
```
yizhong187: Perhaps you would want to wrap this line as it is quite long.

src/main/java/twilight/Twilight.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.ArrayList;
+
+/**
+ * Acts as the main class for running the chatbot.
+ */
+public class Twilight {
+    public static void main(String[] args) {
+        Ui ui = new Ui();
+        Scanner input = new Scanner(System.in);
+        Storage storage = new Storage("data/Twilight.txt");
```
yizhong187: Any reason why you did not abstract out the file path as a constant?

src/main/java/twilight/Task.java
```diff
@@ -0,0 +1,60 @@
...
+     */
+    public String toStorage() {
+        if (isDone) {
+            return "1," + this.description;
+        } else {
+            return "0," + this.description;
+        }
+    }
+
+    public void setDone() {
```
yizhong187: According to the [coding standard](https://se-education.org/guides/conventions/java/intermediate.html): 
> Setter methods for boolean variables must be of the form:
> `void setFound(boolean isFound);`

Perhaps you can lump `setDone` and `setUndone` into one `setDone` method? Or rename the methods to something along the lines of `mark` and `unmark`?

yizhong187 (commented):
I think you can adhere to the [coding standard ](https://se-education.org/guides/conventions/java/intermediate.html) more strictly. Additionally, perhaps packaging some related files together would improve the project structure?

src/main/java/twilight/AddCommand.java
AbdulrahmanAlRammah: I humbly disagree. Since there is an event task this may end up adding confusion as the addcommand class can add any of todo, event, deadline.

src/main/java/twilight/AddCommand.java
AbdulrahmanAlRammah: Yeah I think I should. I just didn't do it thus far out of laziness.

src/main/java/twilight/AddCommand.java
AbdulrahmanAlRammah: For string that are part of a method I think there is more leeway so I will stick to this.

src/main/java/twilight/AddCommand.java
yizhong187: Oh sorry i meant `AddTaskCommand` since `Task` is the parent class of all 3.

src/main/java/twilight/MarkingCommand.java
AbdulrahmanAlRammah: The only problem with this, is that I would have to define the enum in the task class and then check the type of input to convert to an enum and it feels like the enum adds extra steps instead of solving anything, but this is probably because I am unfamiliar with enums and can't see the benefit.

src/main/java/twilight/Deadline.java
AbdulrahmanAlRammah: Agreed.

src/main/java/twilight/TaskList.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    /**
+     * Queries the tasklist for tasks which match and returns a list of them.
+     *
+     * @param q The query.
+     * @return A list of all entries which match the query.
+     */
+    public String query(String q) {
+        TaskList match = new TaskList();
```
ulricolo7: Would it be better if the variable is in plural form? i.e. matches

src/main/java/twilight/Storage.java
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Instantiates Storage.
+     *
+     * @param filePath Path to the storage file from root.
+     */
+    public Storage(String filePath) {
+        this.filePath = filePath;
+    }
+
+    /**
```
ulricolo7, line 27: I might be wrong, but shouldn't there be an empty line after the description and before the @param?

ulricolo7 (commented):
LGTM! All other comments have been mentioned by the other guy haha! Also, good job consistently adding comments for every method!

PR #285, opened 2024-08-27
[DennieDan] iP

DennieDan:
# Tecna
> Bring **magic** to all your tasks 🪄 

* You feel tired seeing the following tasks?
- [ ] Finish Milestone 3
- [ ] Go training at 6 AM
- [ ] Code review
* You don't know how to start a Java project, even a single ``main()`` method as below?
```Java
public class Tecna {
        System.out.println("Hi, I'm Tecna!");
    }
}
```
### It means you _must_ try ***Tecna***

*To know more about Tecna, you can refer to the following pages:*
1. Tecna's [introduction](https://www.winxclub.com/en/article/introducing-myself-i%E2%80%99m-tecna)
2. Tecna's [best moments](https://www.youtube.com/watch?v=7TdIdWjJPuU)


src/main/java/tecna/CommandScanner.java
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * Scans and processes the user's input.
+ * An <code>inputIndex</code> is the index of the corresponding task if the command is <i>mark, unmark, or delete</i>.
+ * An <code>inputTask</code> is the task required to be added via <i>todo, deadline, or event</i> commands.
+ *
+ * @author DennieDan.
+ */
+public class CommandScanner {
+    private final Scanner SCANNER;
```
Gra7ityIC3: Should `SCANNER` be capitalized?

src/main/java/tecna/Storage.java
```diff
@@ -0,0 +1,133 @@
...
+
+    /**
+     * Converts java tecna.Task objects into JSON object and write to a json file.
+     * @param taskList storing the lists of all tasks currently in the app.
+     *
+     * @author Crunchify.com.
+     * https://crunchify.com/how-to-write-json-object-to-file-in-java/.
+     */
+    @SuppressWarnings("unchecked")
+    public void save(TaskList taskList) throws IOException {
```
Gra7ityIC3: Any reason why you chose to save the tasks in JSON instead of plain text?

src/main/java/tecna/TaskList.java
```diff
@@ -0,0 +1,103 @@
+package tecna;
+
+import java.util.ArrayList;
+
+public class TaskList {
+    private ArrayList<Task> tasks;
+    private int size;
```
Gra7ityIC3: Perhaps you could use `tasks.size()` instead of maintaining a separate variable?

src/main/java/tecna/CommandScanner.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    /**
+     * Gets the user's input and parses into the chatbot's valid type command.
+     * @return the corresponding command type in <code>CommandType enum</code>.
+     */
+    public CommandType getRequest() {
+        this.input = this.SCANNER.nextLine().trim();
+        String[] input_words = this.input.split(" ");
+        if (input.equalsIgnoreCase("bye")) {
```
Gra7ityIC3: Perhaps you could use `input_words[0]` for the comparison here (and line 55), ignoring extraneous parameters? That way, even if the user enters `bye 123`, it will be interpreted as `bye`.

Then you could use a `switch` statement which would make your code a lot neater. :smiley:

src/test/java/tecna/CommandScannerTest.java
Gra7ityIC3, line 1: Good job on the adequate test cases! 👍

Gra7ityIC3 (commented):
For your Javadoc comments, there should be an empty line between the description and parameter section.

Also, I think the `this` keyword can be omitted when your variable isn't shadowed by a method or constructor parameter.

src/main/java/tecna/Storage.java
DennieDan: I think JSON has its format with title and value, it would be easier to parse and maintain the validity of the data file also.

src/main/java/tecna/CommandScanner.java
DennieDan: I see what you mean, thanks

src/main/java/tecna/CommandScanner.java
```diff
@@ -0,0 +1,162 @@
...
+     */
+    public String getInput() {
+        return this.input;
+    }
+
+    /**
+     * Gets the user's input and parses into the chatbot's valid type command.
+     * @return the corresponding command type in <code>CommandType enum</code>.
+     */
+    public CommandType getRequest() {
```
PrishaVP: the naming feels a bit vague as it is not really clear what Request means. Especially because the input from the user is referred to as 'input' in other places in your code. Maybe can change the method name to getUserRequest()  or something like that to make it more obvious what the purpose of this method is

src/main/java/tecna/CommandScanner.java
```diff
@@ -0,0 +1,162 @@
...
+        } else {
+            return CommandType.INVALID;
+        }
+    }
+
+    /**
+     * Generates the required index and stores in the <code>inputIndex</code> attribute.
+     * @return <code>true</code> if the parsing process is success and <code>false</code> otherwise.
+     */
+    public boolean generateIndex() {
```
PrishaVP: i like the naming here as it is very clear what the method does

src/main/java/tecna/DialogBox.java
```diff
@@ -0,0 +1,70 @@
...
+import javafx.scene.control.Label;
+import javafx.scene.image.Image;
+import javafx.scene.layout.HBox;
+import javafx.scene.paint.ImagePattern;
+
+/**
+ * Represents a dialog box consisting of an ImageView to represent the speaker's face
+ * and a label containing text from the speaker.
+ */
+public class DialogBox extends HBox {
```
PrishaVP: maybe can fix the spelling of Dialogue (below also). because the naming standards advises full spelling rather than texting-style/casual spelling

src/main/java/tecna/Main.java
```diff
@@ -0,0 +1,37 @@
...
+import javafx.stage.Stage;
+
+/**
+ * A GUI for Tecna using FXML.
+ */
+public class Main extends Application {
+    private Tecna tecna = new Tecna("data/tecna.json");
+
+    @Override
+    public void start(Stage stage) {
```
PrishaVP, line 20: could make the name more precise by naming it startApplication or something like that but start is still quite clear to me so if you feel like putting extra effort then can be more precise i guess

src/main/java/tecna/TaskList.java
```diff
@@ -0,0 +1,112 @@
+package tecna;
+
+import java.util.ArrayList;
+
+public class TaskList {
+    private ArrayList<Task> tasks;
```
PrishaVP: i like that you made it plural

PrishaVP (commented):
naming LGTM except for a few minor suggestions

PR #297, opened 2024-08-27
[seanlim] iP

seanlim:
# Deez

> "Automate the mundane, so you can focus on the extraordinary." – Unknown

**Deez** is your go-to bot for handling the small stuff, freeing up your time for what really matters. It’s:

- Simple to use
- Incredibly efficient
- **Lightning fast**

All you need to do is:

1. Install it from [here](#).
2. Launch the bot.
3. Configure your tasks.
4. Sit back and let Deez handle the rest. 😎

And guess what? It’s **FREE**!

### Features:

- [x] Task management
- [ ] Deadline tracking (coming soon)
- [ ] Custom reminders (coming soon)

---

If you’re into coding, you can use Deez to practice your Java skills too. Check out the `main` method:

```java
public class Main {
        Application.launch(BotApp.class, args);
    }
}
```


src/main/java/deez/Task.java
```diff
@@ -0,0 +1,69 @@
+package deez;
+
+import java.io.Serializable;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Abstract class representing a task.
+ */
+public abstract class Task implements Serializable {
```
Vatinius: If there are no abstract methods being stated, is there still a need for Task to be an abstract class?

src/main/java/deez/Deez.java
```diff
@@ -0,0 +1,231 @@
+package deez;
```
Vatinius, line 1: Would it be better to categorise other classes into more specific packages instead of putting everything in deez?

src/main/java/deez/Parser.java
```diff
@@ -0,0 +1,103 @@
...
+ */
+public class Parser {
+
+    /**
+     * This method parses the given string and returns a pair containing the command and properties.
+     *
+     * @param input The input string to be parsed.
+     * @return A pair containing the command and properties.
+     */
+    static Pair<Command, Properties> parse(String input) {
```
Vatinius, line 28: I like the use of enums to parse a command!

src/main/java/deez/Event.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Returns a string representation of the event.
+     *
+     * @return A string that describes the event and its dates in a human-readable format.
+     */
+    @Override
+    public String toString() {
+        return "[E]"
+            + super.toString()
```
Vatinius: Shouldn't indentation for wrapped lines be 8 spaces more than the parent line?

Vatinius (commented):
Well done! I can see that you've put in effort. Keep an eye out on the small things like following the coding standard that might get you marked down. Keep up the good work and jiayous!

src/main/java/deez/Task.java
seanlim: Hmm, there currently aren't any abstract methods because the descendants don't really share any common non-generic behaviour yet. `Task` should be an abstract class because we don't want to be able to instantiate a `Task` (which is not a `Todo`, nor a `Deadline` or an `Event`) in the main application logic.

src/main/java/deez/Deez.java
seanlim: Yeah, that's a good idea! I'll probably get around to it while refactoring for JavaFX, thanks!

src/main/java/deez/Event.java
seanlim: Yep, thanks for catching this!

src/main/java/deez/TaskList.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.ArrayList;
+
+import deez.tasks.Task;
+
+/**
+ * Represents a list of tasks.
+ */
+public class TaskList implements Serializable {
+
+    private static final long serialVersionUID = 1L;
```
FlyingDonkeys, line 15: Names of `final` constants should be in SCREAMING_SNAKE_CASE, as per CS2103T code quality guidelines 😄

src/main/java/deez/Ui.java
```diff
@@ -0,0 +1,59 @@
...
+        messageConsumer = (String message) -> {
+            System.out.println("WARN: Ui is in headless mode");
+            System.out.println(message);
+        };
+    }
+
+    /**
+     * Initialise Ui with a message consumer
+     */
+    public Ui(Consumer<String> messageConsumer) {
```
FlyingDonkeys, line 27: The JavaDoc for this constructor should explain what the messageConsumer does with a `@params` tag 😄

FlyingDonkeys (commented):
LGTM!

src/main/java/deez/TaskList.java
seanlim: Hi! Actually, the reason why this isn't in screaming snake case is because Java's `Serializable` interface enforces a `serialVersionUID` property that is named in a way that does not follow the CS2103T code standards.

[See info about Serializable here](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/io/Serializable.html). Without a `private static final long` named `serialVersionUID`, the compiler doesn't allow `TaskList` to be `Serializable` 😅

I'm not sure if there is a way around this but I'm open to suggestions!

PR #308, opened 2024-08-27
1

zzzlou:
1


antgyf (changes requested):
LGTM other than a few nits
- consider case if time is omitted from deadline or event, or case where deadline and time format is wrong.

src/main/java/zzbot/DialogBox.java
```diff
@@ -0,0 +1,60 @@
...
+        ObservableList<Node> tmp = FXCollections.observableArrayList(this.getChildren());
+        Collections.reverse(tmp);
+        getChildren().setAll(tmp);
+        setAlignment(Pos.TOP_LEFT);
+    }
+
+    public static DialogBox getUserDialog(String text, Image img) {
+        return new DialogBox(text, img);
+    }
+
```
yooplo, line 64: I would think that javadocs should be added here since it is a public method.

src/main/java/zzbot/Storage.java
```diff
@@ -0,0 +1,66 @@
...
+                arr.add(task);
+            } else {
+                Task task = new ToDos(name, isDone);
+                arr.add(task);
+            }
+        }
+        s.close();
+
+        return arr;
+
```
yooplo: extra line here haha

src/main/java/zzbot/Ui.java
```diff
@@ -0,0 +1,47 @@
+package zzbot;
+
+import java.util.Scanner;
+
```
yooplo, line 4: another extra line here oops

yooplo (commented):
Overall, very well done! I could not find much to nitpick on stylewise. A few small updates suggested.

PR #316, opened 2024-08-27
[DKHB0207] iP

DKHB0207:
# Snipe

> "Your mind is for having ideas, not holding them." – David Allen [source](https://dansilvestre.com/productivity-quotes/)

**Snipe** frees your mind from having to remember things you need to do. It's:

- Text-based
- Easy to learn
- *Super fast* to use

All you need to do is:

1. Download it from [here](https://example.com).
2. Double-click it.
3. Add your tasks.
4. Let it manage your tasks for you. 😊

And it is **FREE**!

## Features

- [x] Managing tasks  
- [ ] Reminders (coming soon)  

```java
public class Main {
    }
}

### CODE REVIEW

File: src/main/java/snipe/core/TaskList.java [hunk 1]
```diff
@@ -0,0 +1,111 @@
...
+        }
+        return tasks.get(index);
+    }
+
+    /**
+     * Returns the number of tasks in the task list.
+     *
+     * @return The size of the task list.
+     */
+    public int size() {
```
sayglenn, line 83: Perhaps using a name like getTaskCount would be more appropriate as methods should be phrased as verbs?

src/test/java/snipe/core/SnipeTest.java
sayglenn, line 1: Are these tests correct? Looks like you might be implementing them in the future, looking forward to it!

src/main/java/snipe/task/Task.java
```diff
@@ -0,0 +1,93 @@
...
+        return this.description;
+    }
+
+    /**
+     * Returns a string representation of the task's status.
+     * A completed task is marked with an 'X', while an incomplete task is marked with a space.
+     *
+     * @return A string representing the task's status.
+     */
+    public String stringStatus() {
```
sayglenn, line 64: This could be changed to an action, such as getStringStatus.

src/main/java/snipe/command/AddCommand.java
```diff
@@ -0,0 +1,44 @@
...
+import snipe.util.Ui;
+
+import java.io.IOException;
+
+/**
+ * The {@code AddCommand} class represents a command to add a new task to the task list.
+ * This command adds the specified task, saves the updated task list to storage, and
+ * provides feedback to the user via the user interface.
+ */
+public class AddCommand extends Command{
```
sayglenn, line 16: nit: missing space between Command and curly brace.

text-ui-test/input.txt
sayglenn, line 1: Perhaps this file could be hidden and only created when user uses the app? If not, the user would have existing tasks on their database already.

src/main/java/snipe/storage/Storage.java
```diff
@@ -0,0 +1,129 @@
...
+    public Storage(String filePath) {
+        this.filePath = Paths.get(filePath);
+    }
+
+    /**
+     * Ensures the directory and file exist, creating them if necessary.
+     *
+     * @throws IOException If an input or output error occurs while creating the directory or file.
+     */
+    public void initialise() throws IOException {
```
sayglenn, line 41: Perhaps the name of this method could be more specific, e.g. initialiseDatabase?

sayglenn (commented):
Hey Daryl! In general, your code looks good and your functions look robust. However, most of your Command classes are missing a space between `... extends Command` and the opening curly brace. I've only commented on the first occurrence of it to reduce the amount of repeated comments, but do take a look at all of the Command files.

Overall, the code is well done and coding standard is generally adhered to. Good job!

PR #319, opened 2024-08-27
[Varun Shukla] iP

Shux347:
**Sam (Super Amazing Machine)**

>Sam is designed to enhance your communication experience by engaging in meaningful and insightful conversations. 

It’s:

•	text-based
•	intuitive to use
•	FAST and RELIABLE

All you need to do is:

1.	Clone the repository from [here](https://github.com/nus-cs2103-AY2425S1/ip/pull/319#). :point_left:
2.	Run the application with your preferred setup.
3.	Interact with Sam to explore its features.

And guess what? It’s FREE!

Features:

- [x] 	 Managing tasks
- [x]  Managing deadlines
- [x]  Managing events
- [ ]  Reminders (coming soon)

If you are a Java programmer, you can use it to practice Java too. Here’s a sample of the `main` method that kicks off Sam:
```java
        new Sam("data/Sam.txt").run();
    }
```


src/main/java/sam/Sam.java
```diff
@@ -0,0 +1,228 @@
...
+    private Storage storage;
+    private Items items;
+
+    /**
+     * Constructs a new instance of the Sam class.
+     *
+     * @param filePath The file path to the storage file.
+     */
+    public Sam() {
+        String filePath = "data/Sam.txt";
```
arnav12344: Should this be a constant?

src/main/java/sam/Event.java
```diff
@@ -0,0 +1,40 @@
...
+     * @param to the end date of the event in the format "dd-MM-yyyy"
+     */
+    public Event(String newname, String from, String to) {
+        super(newname);
+        this.from = LocalDate.parse(from, DATE_FORMAT);
+        this.to = LocalDate.parse(to, DATE_FORMAT);
+    }
+
+    @Override
+    public String toData() {
```
arnav12344, line 28: Should this method name be more specific?

arnav12344 (commented):
I like how clean and well written your code is!

src/main/java/sam/Deadline.java
```diff
@@ -0,0 +1,39 @@
...
+
+    private LocalDate by;
+
+    /**
+     * Constructs a Deadline object with the specified name and deadline.
+     *
+     * @param newname The name of the deadline task.
+     * @param by The deadline of the task in the format "dd-MM-yyyy".
+     */
+    public Deadline(String newname, String by) {
```
RobotWizzard: I didn't like the name "newname" for the parameter here, because:
1. Variable names should be in camelCase (e.g. "newName")
2. I think just "name" is sufficient and self-explanatory for the parameter. Parameters for subclasses don't have to differ from their parent classes.

I noticed the same issue in other classes too.

src/main/java/sam/Deadline.java
```diff
@@ -0,0 +1,39 @@
...
+        super(newname);
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("dd-MM-yyyy");
+        this.by = LocalDate.parse(by, formatter);
```
RobotWizzard: I don't think the variable "formatter" here is necessary, as it is only used in the next line it is declared. Perhaps 
```
this.by = LocalData.parse(by, DateTimeFormatter.ofPattern("dd-MM-yyyy"));
```
is already quite readable.

src/main/java/sam/Event.java
```diff
@@ -0,0 +1,40 @@
+package sam;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Represents an event item that extends the Item class.
+ */
+public class Event extends Item {
+
+    private static final DateTimeFormatter DATE_FORMAT = DateTimeFormatter.ofPattern("dd-MM-yyyy");
```
RobotWizzard, line 10: I like how you stored the date format as a constant here! Maybe you can do the same in the "Deadline" class to make it more consistent.

src/main/java/sam/Item.java
```diff
@@ -0,0 +1,65 @@
...
+     * Marks the item as done.
+     */
+    public void markAsDone() {
+        this.isDone = true;
+    }
+
+    /**
+     * Marks the item as undone.
+     */
+    public void markAsUndone() {
```
RobotWizzard, line 30: I didn't like the name of this method since it is too similar to "markAsDone" to convey that this does the opposite. Furthermore, "undone" is not actually an English word. Perhaps "markAsNotDone" or "unmark" would be better names here.

src/main/java/sam/Items.java
```diff
@@ -0,0 +1,113 @@
+package sam;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Represents a task list.
+ */
+public class Items {
```
RobotWizzard, line 9: Not sure if I like the name "Items" for this class, as it is very similar to the "Item" class. Since the header comment states that this represents a task list, maybe "TasksList" is more appropriate?

src/main/java/sam/Sam.java
RobotWizzard: I disagree with @arnav12344 's comment here. Since it is only used in once in the entire program, I think it can just be hardcoded:
```
storage = new Storage("data/Sam.txt");
```

src/main/java/sam/Sam.java
```diff
@@ -0,0 +1,228 @@
...
+            items = new Items(loadedItems);
+        } catch (IOException e) {
+            ui.showMessage("Error loading tasks. Starting with an empty list.");
+            items = new Items(); // Start with an empty list if loading fails
+        }
+    }
+
+    /**
+     * Handles a single user input and returns the response as a string.
+     * @param input the input from the user
```
RobotWizzard, line 41: Should there be an empty line before this?

src/main/java/sam/Sam.java
```diff
@@ -0,0 +1,228 @@
...
+    public String getResponse(String input) {
+        StringBuilder response = new StringBuilder();
+        
+        try {
+            if (input.equals("bye")) {
+                response.append("Goodbye!\n");
+                if (items.getSize() > 0) {
+                    storage.save(items.getItems());
+                }
+
```
RobotWizzard: I think this is just an accidental empty line. 😳

src/main/java/sam/SamException.java
```diff
@@ -0,0 +1,24 @@
...
+     *
+     * @param message the detail message (which is saved for later retrieval by the getMessage() method)
+     */
+    public SamException(String message) {
+        super(message);
+    }
+
+    /**
+     * Constructs a new SamException with a default detail message.
+     * The default message is "Huh? I'm not quite sure what you mean by that."
```
RobotWizzard, line 19: I like that you included the default message in the header comment. Nice! 👍

src/main/java/sam/Storage.java
```diff
@@ -0,0 +1,96 @@
...
+            case "E":
+                Event event = new Event(name, data[3], data[4]);
+                if (isDone) {
+                    event.markAsDone();
+                }
+                items.add(event);
+                break;
+            default:
+                throw new IOException("Unknown item type in storage file.");
+            }
```
RobotWizzard, line 75: Should this part be indented once more? It is in the switch block.

RobotWizzard (commented):
I thought that your code was really neat and organized 👍, there were just some tiny bits about naming to fix. 😄

PR #323, opened 2024-08-27
[nigellqx] iP

nigellqx:
# Tako

> [!TIP]
> `"The function of good software is to make the complex appear to be simple."` - Grady Booch [source](https://quotefancy.com/quote/1663415/Grady-Booch-The-function-of-good-software-is-to-make-the-complex-appear-to-be-simple)

Tako can manage tasks that you currently have. It is 

- Text-Based
- Easy to use
- ~~Easy~~ VERY EASY to learn

Here are the steps to use it:

1. Download it
2. Add your task
3. Use different features to manage your tasks 
4. **Enjoy it!** 😃 

So here are some of the features:

- [x] _Managing task_
- [x] _Managing task with deadline_
- [x] _Managing task with a date_

```Java
public class Main {
       public static void main (String[] args) {
       }
}
```


src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,286 @@
...
+
+    /**
+     * Checks whether the input command is recognizable by Tako.
+     *
+     * @param command the string that is being input into Tako.
+     * @throws TakoException if the command is not recognized.
+     */
+    public static void parse(String command) {
+        try {
+            if (command.equalsIgnoreCase("bye")) {
```
notnotmax, line 31: I like that you thought of ignoring case when checking. Though, perhaps a switch statement could be easier to read (after preprocessing the command string)?

src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,286 @@
...
+            System.out.println(e.message());
+            Ui.promptInput();
+        }
+    }
+
+
+    /**
+     * Marks the task as complete according to the input
+     *
+     * @param command
```
notnotmax: Minor nitpick - this param seems to be missing a description.

src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    /**
+     * Breaks down the command into its description and when it is due
+     * Make the deadline task according to the description and its due date
+     *
+     * @param command
+     * @throws TakoException if the deadline command is not in the right form
+     */
+    public static void parseDeadline(String command) {
```
notnotmax: The deeply nested code may be difficult to read with really specific string checking - maybe some comments would be valuable?

src/main/java/parser/Parser.java
notnotmax, line 1: Overall it seems like the code works well enough, good job! However, a possible alternative to avoid finicky string matching and deeply nested code could be to make use of the split() method for strings? Is there a reason why the string checking is done using length and charAt?

src/main/java/ui/Ui.java
```diff
@@ -0,0 +1,123 @@
...
+     */
+    public static void greet() {
+
+        System.out.println("Hello! I'm Tako! What can I do for you?\n");
+
+        promptInput();
+    }
+
+    /**
+     * Says goodbye to the user
```
notnotmax: Minor nitpick - maybe add a period after the sentence for consistency with other javadoc method headers?

src/main/java/ui/Ui.java
```diff
@@ -0,0 +1,123 @@
...
+        promptInput();
+    }
+
+    /**
+     * Prints out the message when the user added a task to the list and
+     * ask them for input after that.
+     *
+     * @param task task that is being added to the list.
+     */
+    public static void addTaskMessage(Task task) {
```
notnotmax: Perhaps a name like `printAddTaskMessage` would make it clearer that the method's purpose is to print something, rather than being confused with 'adding a task message'?

src/test/java/parser/ParserTest.java
```diff
@@ -0,0 +1,43 @@
+package parser;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
```
notnotmax: I think explicit imports would better adhere to the coding standard. Maybe this is an unintentional mistake caused by the IDE?

src/test/java/parser/ParserTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+public class ParserTest {
+
+    @Test
+    public void isValidDateTest1() {
```
notnotmax: Minor nitpick - maybe some comments to describe what the test is could be useful? Alternatively, consider the naming convention `unitBeingTested_descriptionOfTestInputs_expectedOutcome` for method names to make them self-explanatory (from week 3 topics).

notnotmax (commented):
lgtm - just a few potential considerations regarding the coding quality.

src/main/java/parser/Parser.java
nigellqx: Did not think about the split() method before. Thank you for your suggestions, really helped me alot 👍

src/main/java/task/TaskList.java
```diff
@@ -0,0 +1,91 @@
...
+        for (int i = 0; i < taskLists.size(); i++) {
+                tasks += taskLists.get(i).toString() + "\n";
+        }
+        return tasks;
+    }
+
+    /**
+     * Helps to add the task from the file when the chatbot just starts.
+     *
+     * @param task task to be loaded into the list
```
estellelim, line 107: Minor coding standard nitpick - punctuation behind each parameter description

estellelim (commented):
LGTM, do not see any other coding standard violations other than the some of the JavaDoc comments missing punctuations or description, and also the wildcard import commented on by @notnotmax
Good job overall, especially on the layout, and variable/ method naming!

PR #324, opened 2024-08-27
[jevan13] iP

jevan13:
# Yapper
> "If it's flipping hamburgers at McDonald's, be the best hamburger flipper in the world. Whatever it is you do you have to master your craft." -Snoop Dogg

Yapper frees your mind of having to remember things you need to do. It's,

text-based
easy to learn
~~FAST~~ **SUPER FAST** to use
All you need to do is,

download it from [here](https://github.com/jevan13/ip/releases/tag/A-Jar).
1. double-click it.
2. add your tasks.
3. let it manage your tasks for you 😉
**And it is FREE!**

Features:
- [x] Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming soon)


src/main/java/Yapper/TaskList.java
```diff
@@ -0,0 +1 @@
...
        }
    }

    /**
     * Exits the program.
     */
    public void exit() {
        System.out.println("Yapper shall yap next time!");
    }
}
```
gnoossk: I like the usage of "to___" for String methods

src/main/java/Yapper/Task.java
```diff
@@ -0,0 +1 @@
...
     * String to be decoded when loading history from the file.
     *
     * @return String representation of the Task object.
     */
    public abstract String toFile();

    public boolean contains(String word) {
        return this.name.contains(word);
    }
}
```
gnoossk: Not sure I like the getter method name for isDone. Can be specified to something like `hasDone` rather than `getDone`

src/main/java/Yapper/Storage.java
thony-ui, line 1: for boolean values such as 
```java
done
```

it should be named as 

```java
isDone
```
for clarity

src/main/java/Yapper/Task.java
thony-ui, line 1: I think the getter method should be

```java
getIsDone
```

because the field that you specified is ```isDone```

src/main/java/Yapper/TaskList.java
thony-ui, line 1: This is where the use of ```Enumeration``` may be useful. Notice how you check for equality in the "if" statements? You can abstract this in the ```Enumeration``` java class and utilise that.

src/test/java/Yapper/TaskListTest.java
thony-ui, line 1: Good use of test cases. However, I would also test for edge cases if it doesn't match the format.

PR #335, opened 2024-08-27
[slye20] iP

slye20:
# Bob 🤓 

> "We love BOB" - Some important people ([source](https://geo-g.fandom.com/wiki/Bob))

Bob helps you to: 
- Keep tracks of your tasks for the *day*
- Manage them

What type of tasks:
1. ~~Task~~ Todos
2. Deadlines
3. Events

```
public class Main {
    }
}
```
`HELLO WORLD`

**Features so far:**
- [x] Tasks
- [ ] GUI


src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,120 @@
...
+public class Parser {
+
+    /**
+     * Parses the input message for marking a task as done and applies the command to the given {@code TaskList}.
+     *
+     * @param message The input message to be parsed (e.g., "mark 1").
+     * @param t The {@code TaskList} where the task will be marked as done.
+     * @throws BobException If the message format is invalid.
+     */
+    public void markTaskParser(String message, TaskList t) throws BobException {
```
JYL27: The formatting of the names of the methods could be somewhat misleading. This method in particular could sound like you are marking a task parser. Perhaps something like parseMarkTaskCommand instead?

I've noticed the same thing in the Ui class as well.

src/main/java/ui/Ui.java
```diff
@@ -0,0 +1,94 @@
...
+     * @param message The message to be sent.
+     */
+    public void sendMessage(String message) {
+        System.out.println("[Bob] " + message);
+    }
+
+    /**
+     * Displays the introductory messages when the application starts.
+     */
+    public void introBobUi() {
```
JYL27: "intro" could come off as somewhat slangy. Maybe just introduceBobUi instead?

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,97 @@
...
+    // The file path where tasks are saved and loaded from
+    static final String FILE_PATH = "data/bob.txt";
+    private Ui ui = new Ui();
+
+    /**
+     * Loads tasks from the specified file into the given {@code TaskList}.
+     *
+     * @param tl The {@code TaskList} where tasks will be loaded into.
+     */
+    public void loadTasks(TaskList tl) {
```
JYL27: I like the method names here, very clear and concise 👍

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,97 @@
...
+        File f = new File(FILE_PATH);
+        try {
+            Scanner s1 = new Scanner(f);
+            while (s1.hasNext()) {
+                String t = s1.nextLine();
+                String[] task_list = t.trim().split(" \\| ");
+                Task x;
```
JYL27: I like the general consistency of your abbreviations. This particular Task object is abbreviated as x instead of t like in other places though. Any reason why the String object is abbreviated as t and the Task object is abbreviated as x, rather than abbreviating the task object as t and using another abbreviation for the String object?

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,97 @@
...
+     *
+     * @param tl The {@code TaskList} where tasks will be loaded into.
+     */
+    public void loadTasks(TaskList tl) {
+        File f = new File(FILE_PATH);
+        try {
+            Scanner s1 = new Scanner(f);
+            while (s1.hasNext()) {
+                String t = s1.nextLine();
+                String[] task_list = t.trim().split(" \\| ");
```
JYL27: Not sure I like the usage of snake case here. Any reason why this variable in particular is in snake case instead of camel case?

JYL27 (commented):
LGTM👍. Just a few nits regarding variable and method naming.

src/main/java/task/Event.java
```diff
@@ -0,0 +1,37 @@
...
+        return "[E]" + super.toString() + " (from: " + starting + " to: " + ending + ")";
+    }
+
+    /**
+     * Formats event object into a string to be saved in.
+     * @return String to be saved.
+     */
+    @Override
+    public String toSave() {
+        return "E" + " | " + (this.isDone ? 1 : 0) + " | " + this.description + " | " + this.start + " | " + this.end;
```
matthewyip1511: This line is a bit long, you might want to add a line break here 🙂

src/main/java/task/TaskList.java
```diff
@@ -0,0 +1,135 @@
...
+            }
+        }
+    }
+
+    /**
+     * Marks a task as done based on its position in the task list.
+     *
+     * @param num The index of the task to be marked as done (0-based index).
+     */
+    public void markTask(int num) {
```
matthewyip1511: this might be better named as `markTaskAsDone` to be more descriptive 🙂

src/main/java/task/TaskList.java
```diff
@@ -0,0 +1,135 @@
...
+        t.markAsDone();
+        this.ui.markTaskUi(t);
+    }
+
+    /**
+     * Unmarks a task as not done based on its position in the task list.
+     *
+     * @param num The index of the task to be unmarked as not done (0-based index).
+     */
+    public void unmarkTask(int num) {
```
matthewyip1511: similarly this could be `unmarkTaskAsDone` or `markTaskAsUndone`

src/main/java/ui/Ui.java
matthewyip1511: I agree, `introduce` would be clearer in this context

src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,120 @@
...
+     *
+     * @param message The input message to be parsed (e.g., "todo read book").
+     * @param t The {@code TaskList} where the ToDo task will be added.
+     * @throws BobException If the message format is invalid or the description is empty.
+     */
+    public void addToDoParser(String message, TaskList t) throws BobException {
+        if (!message.matches("^todo.*")) {
+            throw new BobException("Invalid format");
+        }
+        String x = message.replaceFirst("todo ", "");
```
matthewyip1511: The variable name should be more descriptive here 🙂 , instead of `x` it could be `parsedMessage`

src/main/java/task/Task.java
```diff
@@ -0,0 +1,58 @@
...
+    protected String description;
+    protected boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    /**
+     * Get status icon
```
matthewyip1511: Nitpicking here, but it should be `Gets` rather than `Get`. Otherwise most of the javadoc comments are in line with convention, good job!

src/main/java/Bob.java
```diff
@@ -0,0 +1,103 @@
...
+                this.parser.addToDoParser(message, this.tasks);
+                break;
+            case "deadline":
+                this.parser.addDeadlineParser(message, this.tasks);
+                break;
+            case "event":
+                this.parser.addEventParser(message, this.tasks);
+                break;
+            case "find":
+                this.parser.findParser(message, this.tasks);
```
matthewyip1511: Function names here are clear and easy to follow 👍

matthewyip1511 (commented):
Coding standards are mostly met overall! Just a few nits here and there, but otherwise good work so far 👍

PR #371, opened 2024-08-28
[siriousguy] iP

siriousguy:
### Handsome

> "Do not take life too seriously. You will never get out of it alive.” - Elbert Hubbard [source](https://www.adobe.com/express/discover/quotes/funny)

Handsome helps you with your life. It is

- user friendly
- ~fun~ **SUPER FUN** to talk to 🤠
- REALLY QUICK to learn

All you need to do is,

1. download it.
2. double-click the file.
3. add your tasks.
4. sit back as you enjoy more free time 🕰️

Features:
- [x]  Managing tasks
- [ ]  Managing deadlines (coming soon)
- [ ] Managing events (coming soon)
- [ ]  Sending reminders (coming soon)

Well, here's the `main` method if you would like to check it out:

```
        new Handsome("./data/handsome.txt").run();
    }
```


src/main/java/Handsome.java
```diff
@@ -0,0 +1,153 @@
...
+        System.out.println("Hey, here are the tasks you have!");
+        for (int i = 0; i < data.size(); i++) {
+            System.out.println((i + 1) + "." + data.get(i).toString());
+        }
+    }
+    private static void markDone(int taskNum) {
+        data.get(taskNum).markAsDone();
+        System.out.println("Congrats, you finally finished this task!");
+        System.out.println(data.get(taskNum).toString());
+    }
```
tansanchian: Perhaps it would be better if there is a line break between these methods?

src/main/java/Handsome.java
```diff
@@ -0,0 +1,153 @@
...
+
+    private static void markUndone(int taskNum) {
+        data.get(taskNum).markAsUndone();
+        System.out.println("Hey, the task is not done but time to stop procrastinating?");
+        System.out.println(data.get(taskNum).toString());
+    }
+
```
tansanchian, line 70: There is a line break here, which is inconsistent with how you deal with line breaks above. Any reason why you did it this way?

src/main/java/Handsome.java
```diff
@@ -0,0 +1,153 @@
...
+    private static void deleteCheck(String[] des) throws HandsomeException {
+        if (des.length < 2 ) {
+            throw new HandsomeException("Hey, please let me know the task that you want to delete.");
+        }
+        int taskNum = Integer.parseInt(des[1]) - 1;
+        if (taskNum > data.size()) {
+            throw new HandsomeException("Hey, that task simply... does not exist.");
+        }
+        // temporarily store task so i can let user know what is deleted
```
tansanchian: I like how you give comments inside the method body.

src/main/java/Handsome.java
```diff
@@ -0,0 +1,153 @@
...
+
+import task.Task;
+import task.ToDo;
+/**
+ * Chatbot that is able to differentiate tasks
+ */
+public class Handsome {
+    private static final ArrayList<Task> data = new ArrayList<>();
+    public static class Deadline extends Task {
+        protected String by;
```
tansanchian: I like how the protected access modifier is used here.

tansanchian (commented):
LGTM. Overall is great!

src/main/java/Handsome.java
```diff
@@ -0,0 +1,77 @@
...
+        ui.welcomeUser();
+        boolean isExit = false;
+        while(!isExit) {
+            String fullCommand = ui.readCommand();
+            try {
+                Command c = Parser.parse(fullCommand);
+                c.execute(tasks, ui, storage);
+                isExit = c.isExit();
+            } catch (IOException | HandsomeException e) {
+                throw new RuntimeException(e);
```
kushaldas27, line 57: Instead of throwing RuntimeExceptions that would simply terminate the program, consider ```e.printStackTrace()``` or ```showError(e.getMessage())``` that would log the error or provide feedback to the user.

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,121 @@
...
+            storedTask.append(" | ").append(DESIRED_FORMAT.format(((Deadline) task).by));
+        } else if (task instanceof Event) {
+            storedTask.append("E");
+            storedTask.append(" | ").append(task.isDone ? "1" : "0");
+            storedTask.append(" | ").append(task.description);
+            storedTask.append(" | ").append(DESIRED_FORMAT.format(((Event) task).from));
+            storedTask.append(" | ").append(DESIRED_FORMAT.format(((Event) task).to));
+        }
+        return storedTask.toString();
+    }
```
kushaldas27: Consider creating an abstract ```toFileString()``` in your Task class and override the method in your ToDo, Deadline and Event classes depending on the format required. Would help with repeated code and maintaining your code if you add new types of Tasks in the future.

kushaldas27 (commented):
Good Job 👍

PR #374, opened 2024-08-28
[teamotitan] iP

teamotitan:
# TimoPro
> #### "Life is a paradise of endless possibilities!" - [Mehmet Murat ildan](https://quotefancy.com/quote/2348216/Mehmet-Murat-ildan-Life-is-a-paradise-of-endless-possibilities)

TimoPro frees your mind of having to remember to many things that you need to do every single day! It's,
- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,
1) download it from [here](https://github.com/teamotitan/ip/releases/tag/A-Jar)
2) double-click it.
3) add your tasks.
4) let it manage your tasks for you! 👍 

And it is **FREE!**

Features:
- [x] Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming soon)

If you are a Java programmer, you can use it to practice Java too. Here's the ```main``` method

```java
public class Main {
    }
}
```


src/main/java/Timo/Timo.java
```diff
@@ -0,0 +1,584 @@
...
+import java.time.DateTimeException;
+import java.time.LocalDateTime;
+import java.util.Scanner;
+import java.util.*;
+import java.io.File;
+import java.time.format.DateTimeFormatter;
+
+class Task {
+    private String taskype = "T";
+    private boolean mark;
```
rayray39: Is there a typo in ```String taskype```? Perhaps ```boolean mark``` could sound more like a boolean, for example adding a prefix such as is, can, has etc.

src/main/java/Timo/Timo.java
```diff
@@ -0,0 +1,584 @@
...
+    private String val;
+    public Task(boolean mark, String val) {
```
rayray39: Perhaps something more descriptive could be used in place of ```val```?

src/main/java/Timo/Timo.java
```diff
@@ -0,0 +1,584 @@
...
+                String tmp = s.nextLine();
+                if (tmp.startsWith("[T]")) {
+                    String[] a = tmp.split("] ", 2);
+                    if (Character.compare(tmp.charAt(4), 'X') == 0) {
+                        arr.add(new Todo(true, a[1]));
+                    } else {
+                        arr.add(new Todo(false, a[1]));
+                    }
+                } else if (tmp.startsWith("[D]")) {
+                    //remove the [D][?] from the line
```
rayray39: I like how comments are added to facilitate understanding how the code creates a deadline and event, perhaps similar comments could be added into todo as well?

src/main/java/Timo/Timo.java
```diff
@@ -0,0 +1,584 @@
...
+                this.storage.store(this.taskList.showList());
+                break;
+
+            case "list":
+                this.ui.printList(this.taskList);
+                break;
+
+            default:
+                String s = String.valueOf(command.charAt(command.length() - 1));
+                if (command.startsWith("mark")) {
```
rayray39: Any reason why mark, unmark, delete, todo, deadline, event and find are not cases of the switch block?

src/main/java/Timo/Timo.java
```diff
@@ -0,0 +1,584 @@
...
+    }
+
+    public void run() {
+        //welcome
+        ui.greet();
+
+        //print list initially
+        ui.printList(this.tasks);
+
+        boolean isExit = false;
```
rayray39: Good use of a boolean variable that sounds like booleans

rayray39 (commented):
Overall the code was easy to understand and has a logical flow. Consider splitting the classes into different files if that makes sense.

src/main/java/Timo/Timo.java
teamotitan: Okay thanks for pointing out my mistakes. I will go and resolve them!

src/main/java/Timo/Timo.java
teamotitan: Ok that's a good idea, thanks!

src/main/java/Timo/Timo.java
teamotitan: Ok I will do that, thanks!

src/main/java/Timo/Timo.java
teamotitan: You are right, I will go edit it!

src/main/java/Timo/Timo.java
teamotitan: Thank you!

src/main/java/Timo/Timo.java
```diff
@@ -0,0 +1,584 @@
...
+
+class TimoException extends Exception {
+
+    public TimoException(String errorMessage) {
+        super(errorMessage);
+    }
+
+}
+
+//Timo.Storage: beginning I read from the file, end I update the file
```
yanqiyqh: Maybe this can be worded better

src/main/java/Timo/Timo.java
```diff
@@ -0,0 +1,584 @@
...
+
+    public String getTask() {
+        return this.tasktype;
+    }
+
+    public String getStatusIcon() {
+        return (this.mark ? "X" : " ");
+    }
+
+    @Override
```
yanqiyqh: Nice that you included @Override tags

src/main/java/Timo/Timo.java
```diff
@@ -0,0 +1,584 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+import java.time.DateTimeException;
+import java.time.LocalDateTime;
+import java.util.Scanner;
+import java.util.*;
+import java.io.File;
+import java.time.format.DateTimeFormatter;
+
+class Task {
```
yanqiyqh: Consider separating the classes into different files? e.g. Task.java, Todo.java etc

src/main/java/Timo/Timo.java
```diff
@@ -0,0 +1,584 @@
...
+        return "[" + this.tasktype + "]" + super.toString() +
+                " (by: " + this.date.format(DateTimeFormatter.ofPattern("MMM dd yyyy HHmm"))
+                + ")";
+    }
+}
+
+class Event extends Task {
+    private String from;
+    private String to;
+    private String tasktype = "E";
```
yanqiyqh: Since tasktype is tied to the Class and does not change, it should be a static variable: `private static String TASKTYPE = "E";`

yanqiyqh (commented):
Overall, your code largely follows the coding standards. Nice!

PR #409, opened 2024-08-28
[Gra7ityIC3] iP

Gra7ityIC3:
MorganaPro frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- ~~FAST~~ _SUPER_ FAST to use

All you need to do is,

1. download it from [here](https://github.com/Gra7ityIC3/ip).
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you :smile_cat:

And it is **FREE**!

Features:

- [x] Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming soon)

---

```java
public class Main {
    }
}
```


src/main/java/morgana/task/TaskList.java
```diff
@@ -0,0 +1,64 @@
+package morgana.task;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Represents a list of tasks, providing methods to add, remove, and
+ * retrieve tasks from the list.
+ */
+public class TaskList {
```
HanB1n: Would it be better if the method names were more descriptive eg. addTask, getTask, removeTask, instead of simply add, get, remove?

src/main/java/morgana/exceptions/MorganaException.java
```diff
@@ -0,0 +1,17 @@
...
+ * This exception is thrown when specific errors related to the application's
+ * functionality occur.
+ */
+public class MorganaException extends Exception {
+    /**
+     * Constructs a {@code MorganaException} with the specified detail message.
+     *
+     * @param message The detail message describing the specific error that occurred.
+     */
+    public MorganaException(String message) {
```
HanB1n, line 14: Would the code be more readable if different error types are extracted?

HanB1n (commented):
The code was well written and easy to understand with insignificant coding standard violations. Good Job!

src/main/java/morgana/Morgana.java
```diff
@@ -0,0 +1,63 @@
...
+import morgana.storage.Storage;
+import morgana.task.TaskList;
+import morgana.ui.Ui;
+
+/**
+ * Entry point of the Morgana application.
+ * Initializes the application and starts the interaction with the user.
+ */
+public class Morgana {
+    private static final String NAME = "Morgana";
```
leroytan: I like how you used caps for a constant

src/main/java/morgana/commands/AddCommand.java
```diff
@@ -0,0 +1,46 @@
...
+import morgana.task.TaskList;
+import morgana.ui.Ui;
+
+/**
+ * Represents an abstract command to add a task to the task list.
+ * Concrete subclasses should implement the {@link #createTask(String)} method
+ * to define how the task is created based on the user's input.
+ */
+public abstract class AddCommand extends Command {
+    private final String args;
```
leroytan, line 20: Should this be in caps instead since it is a final variable?

src/main/java/morgana/commands/FindCommand.java
```diff
@@ -0,0 +1,38 @@
...
+import morgana.storage.Storage;
+import morgana.task.Task;
+import morgana.task.TaskList;
+import morgana.ui.Ui;
+
+/**
+ * Represents a command to find tasks that match a given keyword in their description.
+ */
+public class FindCommand extends Command {
+    private final String keyword;
```
leroytan, line 15: Consider changing the keyword to CAPS, I have also seen the issue elsewhere in the code

src/test/java/morgana/DeadlineCommandTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+import morgana.commands.DeadlineCommand;
+import morgana.exceptions.MorganaException;
+import morgana.storage.Storage;
+import morgana.task.TaskList;
+import morgana.ui.Ui;
+
+public class DeadlineCommandTest {
+    @Test
+    public void deadlineCommand_validInput_success() throws MorganaException {
```
leroytan: I like how u named the testcases

leroytan (commented):
LGTM, just need to change some final variable to CAPS

src/main/java/morgana/commands/AddCommand.java
Gra7ityIC3: `args` is not a constant so it does not need to be capitalized.

src/main/java/morgana/exceptions/MorganaException.java
Gra7ityIC3: You're right. Thank you for the suggestion.

PR #415, opened 2024-08-28
[Yaswitha Nalluri] iP

yastsc:
## Star ⭐️

Star is a chatbot that manages your life for you so that YOU can live up to your ⭐️ potential! It can do SO much:
- [ ] manage your deadlines
- [ ] manage your todos
- [ ] manage your events

And there's more! It is incredibly fun to use because its:
* text-based
* simple to learn
* fast to use

Start using Star NOW!
1. download it from [here]()
2. double-click it
3. add your tasks
4. sit back and watch it arrange everything for you 🤩
> Star, star, star, star, star, star - Megan Thee Stallion

You can also use this application to practise Java! Here's the main method:
```
public class Main {
        Application.launch(Star.class, args);
    }
}
```


src/main/java/star/command/Command.java
```diff
@@ -0,0 +1,33 @@
...
+     * Executes the set of actions within.
+     *
+     * @param tasks   which is the task list on which the command is executed.
+     * @param ui      which is the UI that is used to interact with the user.
+     * @param storage which is the storage used to load and save tasks.
+     * @throws StarException which is an Exception specific to Star that is thrown when there is an error.
+     */
+    public String execute(TaskList tasks, Ui ui, Storage storage) throws StarException {
+        return "";
+    }
```
EyuGongYi: Small thing to nitpick, but maybe you can change this to an abstract class? As execute dont really do anything and was thinking potentially down the line something might become an issue?

src/main/java/star/main/Storage.java
```diff
@@ -0,0 +1,77 @@
...
+                switch (inputs[0]) {
+                case "T":
+                    task = new ToDo(inputs[2]);
+                    break;
+                case "D":
+                    task = new Deadline(inputs[2], inputs[3]);
+                    break;
+                case "E":
+                    task = new Event(inputs[2], inputs[3]);
+                    break;
```
EyuGongYi: Nice use of switch , just thinking that you could add enum for the Task type in the case that you decide to change the icon for each of the task.

src/main/java/star/task/Deadline.java
```diff
@@ -0,0 +1,67 @@
...
+//    public String getTask() {
+//        String[] strArray = super.getDescription().split("/");
+//        String[] startArr = strArray[0].split(" ", 2);
+//        return startArr[1];
+//    }
+
+    @Override
+    public String getOutput() {
+        if (by == null) {
+            DateTimeFormatter formatter = DateTimeFormatter.ofPattern("MMM d yyyy HH:mm");
```
EyuGongYi, line 59: Could this be extracted out? so that you dont have to redeclare them constantly and you dont have to search through the code to change each and everyone of it.

src/main/java/star/exception/StarException.java
```diff
@@ -0,0 +1,48 @@
+package star.exception;
+
+/**
+ * Represents exceptions that are specific to the Star chatbot.
+ * Inherits from the Exception class.
+ */
+public class StarException extends Exception {
```
EyuGongYi, line 7: Like the way you do the different exceptions. Its quite an interesting way to handle it

EyuGongYi (commented):
Really love how the code is organised overall other than the things i commented. Good job on the code standards too

PR #422, opened 2024-08-28
[Mau Ze Wei] iP

wassilililily:
# BottleOpener

> "I can't believe you lost our bottle opener." - J.D. 

Don't worry about losing your bottle opener again (or anything else for that matter) as BottleOpener will
* help you keep track of tasks
* search through your list of tasks quickly

All you need to do is:
1. Download it from [here](https://github.com/wassilililily/ip/releases/tag/A-Jar)
2. Run it
3.  **Profit?!?** 👍 

Some of the basic commands are:
```
bye
list
delete
todo
deadline
event
```

You can even do this:
- [ ] mark
- [ ] unmark

Here is a snippet of the code:

```ruby
public class BottleOpener {
        String botName = "BottleOpener";
```


src/main/java/BottleOpener/BottleOpener.java
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * The main method of the BottleOpener.BottleOpener chatbot. It sets up the user interface, loads tasks from storage,
+     * and starts the command processing loop.
+     *
+     * @param args Command-line arguments (not used).
+     */
+    public static void main(String[] args) {
+        String botName = "BottleOpener";
+        String filepath = "data/BottleOpener.txt";
```
hanscau: Should change these into a constant instead since they are hard coded
```java
private final String BOT_NAME = "BottleOpener";
private final String FILE_PATH = "data/BottleOpener.txt";
```

src/main/java/BottleOpener/Deadline.java
```diff
@@ -0,0 +1,134 @@
...
+     * The deadline is parsed from the given string using the defined {@link #FORMATTER}.
+     * </p>
+     *
+     * @param description The description of the BottleOpener.Deadline task.
+     * @param deadline    The deadline of the task as a string.
+     */
+    public Deadline(String description, String deadline) {
+        super(description);
+        try {
+            this.deadline = LocalDateTime.parse(deadline, FORMATTER);
```
hanscau: Very cool to allow different formats of datetime 👍

src/main/java/BottleOpener/Event.java
```diff
@@ -0,0 +1,161 @@
...
+                .appendOptional(DateTimeFormatter.ofPattern("MM/dd/yyyy HHmm"))
+                .appendOptional(DateTimeFormatter.ofPattern("MM/dd/yyyy HH:mm"))
+                .appendOptional(DateTimeFormatter.ofPattern("MM/dd/yyyy"))
+                .appendOptional(DateTimeFormatter.ofPattern("MM-dd-yyyy HHmm"))
+                .appendOptional(DateTimeFormatter.ofPattern("MM-dd-yyyy HH:mm"))
+                .appendOptional(DateTimeFormatter.ofPattern("MM-dd-yyyy"))
+                .parseDefaulting(ChronoField.HOUR_OF_DAY, 0)
+                .parseDefaulting(ChronoField.MINUTE_OF_HOUR, 0)
+                .toFormatter();
+    }
```
hanscau: Duplicated code, maybe can abstract them away? or just have one of the class import it from the other class

src/main/java/BottleOpener/Event.java
```diff
@@ -0,0 +1,161 @@
...
+    public Event(String description, boolean status, String start, String end) {
+        super(description, status);
+        try {
+            this.start = LocalDateTime.parse(start, FORMATTER);
+            this.end = LocalDateTime.parse(end, FORMATTER);
+        } catch (DateTimeParseException e) {
+            System.out.println("Invalid event format!");
+            this.start = LocalDateTime.now();
+            this.end = LocalDateTime.now();
+        }
```
hanscau: nit: does not seem right to force the event to have current time for start and end if the format is wrong

src/main/java/BottleOpener/Storage.java
```diff
@@ -0,0 +1,85 @@
...
+                        String type = word[0];
+                        boolean status = word[1].contains("X");
+
+                        if (type.equals("T")) {
+                            Task newTask = new ToDo(word[2], status);
+                            tasks.addTask(newTask);
+                        } else if (type.equals("D")) {
+                            Task newTask = new Deadline(word[2], status, word[3]);
+                            tasks.addTask(newTask);
+                        } else if (type.equals("E")) {
```
hanscau: nit: can call `Task.getType()`, `Deadline.getType()`, and `Event.getType()` instead so there is no need to hardcode it in and since you have them defined in the individual classes already

hanscau (changes requested):
Overall very good and clean code writing. Very verbose javadocs and love the fact that the events are treated as if they are immutable. 

Just a few suggestions and weird behaviour that I don't think make sense for the user

src/main/java/BottleOpener/BottleOpener.java
```diff
@@ -0,0 +1,47 @@
...
+        Ui ui = new Ui(botName);
+        Storage storage = new Storage(filepath);
+
+        System.out.println(ui.showGreeting());
+
+        BufferedReader br = new BufferedReader(new InputStreamReader(System.in));
+
+        Tasklist tasks = new Tasklist();
+        tasks = storage.load(tasks);
+        boolean flag = false;
```
PresleyChew: Perhaps rename "flag" to something else, as booleans should sound like booleans, but overall code is amazing! 😸

src/main/java/BottleOpener/BottleOpener.java
```diff
@@ -0,0 +1,47 @@
...
+        while (!flag) {
+            storage.save(tasks);
+            String inp = "";
```
PresleyChew: Perhaps consider renaming "inp" to something more clear!

src/main/java/BottleOpener/ToDo.java
```diff
@@ -0,0 +1,57 @@
...
+     */
+    public ToDo(String description, boolean status) {
+        super(description,status);
+    }
+
+    public ToDo markAsDone() {
+        return new ToDo(this.description, true);
+    }
+
+    public ToDo markAsUndone() {
```
PresleyChew: Great code naming! Using verbs for actions, great job ! 🔥

src/main/java/BottleOpener/Tasklist.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.ArrayList;
+
+/**
+ * The {@code BottleOpener.Tasklist} class manages a list of tasks, allowing operations such as adding,
+ * deleting, marking, and unmarking tasks. It provides functionality to interact with the
+ * task list and retrieve information about the tasks.
+ */
+public class Tasklist {
+    public int index;
+    public ArrayList<Task> tasklist;
```
PresleyChew: Perhaps consider renaming it to taskLists, following camelCasing as well as distinguishing clearly between single-valued and multi-valued variables! Small nit picking!

PresleyChew (commented):
LGTM with minor nit picking with naming, otherwise code is great! 💯

PR #429, opened 2024-08-28
[KLinkesV] iP

LinkesV:
## Cypher Chat Bot 

> Any fool can write code that a computer can understand. Good programmers write code that humans can understand. --Martin Fowler [[source](http://www.amazon.com/exec/obidos/redirect?link_code=as2&path=ASIN/0201485672&tag=anothedayinmy-20&camp=1789&creative=9325)]

### Why use **Cypher**?:
- Easy to learn
- Use CLI for all commands
- ~~Fast~~ Faster to use than other ChatBots out there

Features of Cypher:
- [x] Manage different types of task such as Todo, Deadlines and Events
- [ ]  Mark task as complete
- [ ] Store previous session data in a local file
- [x] Easy to use text-based commands

**Download** and use now by follow these steps:
1. download it from here
2. run the jar file in terminal
3. and you are good to go 

Here is some sample code from this project:
```
public class Main {
    }
}

```

.


src/main/java/cypherchatbot/Cypher.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /**
+     *  Main event loop of the Cypher Chat Bot Application.
+     *  Event loops runs until the last command given has an exit status of true
+     */
+
+    public void run() {
+        ui.greet();
+        boolean shouldEnd = false;
```
annabellekk: Would it be better to use a name such as 'hasEnded' here instead?

src/main/java/cypherchatbot/command/ByeCommand.java
```diff
@@ -0,0 +1,16 @@
...
+import cypherchatbot.util.TaskList;
+import cypherchatbot.util.Ui;
+
+public class ByeCommand extends Command {
+
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        ui.goodBye();
+    }
+
+    public boolean isExit() {
```
annabellekk: Perhaps could change this to 'returnExitStatus' to make it a verb

src/main/java/cypherchatbot/command/DeadlineCommand.java
```diff
@@ -0,0 +1,60 @@
...
+import cypherchatbot.util.TaskList;
+import cypherchatbot.util.Ui;
+
+
+/**
+ * The Deadline class represents a command to create a new Deadline task
+ * and add it to the task list in the Cypher chat bot application.
+ */
+public class DeadlineCommand extends Command {
+    private String[] command;
```
annabellekk, line 20: Can change it to 'commands' instead since it is a collection of String! :)

src/main/java/cypherchatbot/task/Task.java
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    @Override
+    public String toString() {
+        String str = this.completed ? "[X] " : "[ ] ";
+        str += this.description;
+        return str;
+    }
+
+    public abstract String toStringinFile();
```
annabellekk: Could change the name to 'toStringInFile()' to follow the camelCase convention

src/main/java/cypherchatbot/util/CommandReader.java
```diff
@@ -0,0 +1,101 @@
...
+            case UNMARK:
+                // Need check if that is number
+                int unmarkVal = Integer.parseInt(command[1]) - 1;
+                return new UnmarkCommand(unmarkVal);
+            case BYE:
+                return new ByeCommand();
+            case HELP:
+                return new HelpCommand();
+            case DELETE:
+                int delVal = Integer.parseInt(command[1]) - 1;
```
annabellekk, line 61: Perhaps could expand on the name of the variable delVal to improve readability

annabellekk (commented):
Good implementation overall! Just some comments on minor naming improvements. I like how you abstracted out all the commands, tasks, and util into not just different classes but different packages as well which makes the structure clearer. Good job!!

src/main/java/cypherchatbot/Cypher.java
mongj: I think `shouldEnd` works! `hasEnded` implies that the program has already ended which might be counterintuitive.

src/main/java/cypherchatbot/Cypher.java
```diff
@@ -0,0 +1,92 @@
...
+                ui.showError(e.getMessage());
+            }
+        }
+    }
+
+    /**
+     *  Initializes the Cypher class with the preferred file path. Executes the run
+     *  method in order to start up the Cypher Chat Bot Application
+     */
+
```
mongj: Could remove the linebreak here else it is not clear that the comment above is meant for the `main` method.

```suggestion

```

src/main/java/cypherchatbot/util/CommandReader.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /**
+     * Parses through the given input by user and instantiates a new Command Class through the use
+     * of enums and switch case statements.
+     *
+     *
+     * @param input The string input given by the user
+     * @return A Command Object depending on the input given by the user
+     * @throws CypherException Throws a custom exception if an invalid input is given by the user
```
mongj: I think it is not necessary to mention "Throws a custom exception" as it is implied. 

```suggestion
     * @throws CypherException If an invalid input is given by the user.
```

I believe this is more closely aligned with the Javadoc conventions as well.

src/main/java/cypherchatbot/util/CommandReader.java
```diff
@@ -0,0 +1,101 @@
...
+
+        LIST, TODO, EVENT, DEADLINE, MARK, UNMARK, BYE, HELP, DELETE, FIND
+
+    }
+
+    /**
+     * Parses through the given input by user and instantiates a new Command Class through the use
+     * of enums and switch case statements.
+     *
+     *
```
mongj, line 31: Just one empty line is required

```suggestion
     *
```

src/main/java/cypherchatbot/util/CommandReader.java
```diff
@@ -0,0 +1,101 @@
...
+public class CommandReader {
+
+    private static enum Commands {
+
+        LIST, TODO, EVENT, DEADLINE, MARK, UNMARK, BYE, HELP, DELETE, FIND
+
+    }
+
+    /**
+     * Parses through the given input by user and instantiates a new Command Class through the use
```
mongj: It may not be necessary to mention that the function uses "enums and switch case statements", since the purpose of comments is to help explain your abstraction to people such that they don't have to read the actual code.

Personally, I think it is sufficient to say something like "Parses through the given input by user and instantiates a new Command Class". Just a thought!

src/main/java/cypherchatbot/util/CommandReader.java
```diff
@@ -0,0 +1,101 @@
...
+     * of enums and switch case statements.
+     *
+     *
+     * @param input The string input given by the user
```
mongj, line 32: Remember to add a period after each line, i.e.

```suggestion
     * @param input The string input given by the user.
```

Same for other places

src/main/java/cypherchatbot/task/Task.java
mongj: Agreed, remember to change it in other places in the code too :)

src/main/java/cypherchatbot/task/Task.java
```diff
@@ -0,0 +1,28 @@
...
+public abstract class Task {
+    protected String description;
+    protected Boolean completed;
+
+    public Task(String desc) {
+        this.description = desc.trim();
+        this.completed = false;
+    }
+
+    public void completeTask() {
```
mongj: Was wondering if it might be cleaner to name this as `complete()` instead? Calling `completeTask` on a `Task` object might be redundant.

src/main/java/cypherchatbot/task/Task.java
```diff
@@ -0,0 +1,28 @@
...
+        this.completed = true;
+    }
+
+    public void incompleteTask() {
```
mongj: `incompleteTask()` seems a tad awkward for a method name as incomplete is not a verb. Could possibly consider alternatives like `markAsIncomplete()`

src/main/java/cypherchatbot/util/Ui.java
```diff
@@ -0,0 +1,112 @@
+package cypherchatbot.util;
+
+import java.util.Scanner;
+
+import cypherchatbot.Cypher;
+import javafx.animation.PauseTransition;
+import javafx.application.Platform;
+import javafx.util.Duration;
+
+/**
```
mongj, line 8: I like the descriptive class comments throughout the code!

mongj (commented):
LGTM! Not much to fix wrt coding standards 👍

PR #435, opened 2024-08-28
[Bhudisanont Kanit] iP

JumpyJay:
# Milo
> Tracking task made __simple__ (_and ~~fast~~ **REALLY FAST**_ 💨 💨 ) 
- [ ] like ticking this checkbox  ✅ 

With milo you could 
- Track to-do task 🏙️  
- Track deadlines 🌃 
- Track events 🌠 
- And more ❗  
.
.

To get started, all you have to do is
1. Download it from [here](https://github.com/JumpyJay/ip/releases/tag/A-Jar)
2. Start up the Jar file
3. Let Milo `do` the rest 😸😸😸

```
function tryNow() {
  console.log("Give it a go—it's absolutely free!");
}
```


src/test/java/Milo/MiloTest.java
```diff
@@ -0,0 +1,40 @@
+package Milo;
+import org.junit.jupiter.api.Test;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.PrintStream;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MiloTest {
+    @Test
+    public void testGreetingAndExit() {
```
Edwin1022: This works, but you could also try something like greetingAndExit_success?

src/test/java/Milo/UiTest.java
```diff
@@ -0,0 +1,26 @@
+package Milo;
+import org.junit.jupiter.api.Test;
+import java.io.ByteArrayOutputStream;
+import java.io.PrintStream;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class UiTest {
+    @Test
+    public void testGreetUser() {
```
Edwin1022: This works, but you could also try something like greetUser_success?

src/main/java/Milo/TaskObj/TaskType.java
```diff
@@ -0,0 +1,15 @@
+package Milo.TaskObj;
+
+/*
+ * Represents the different possible task type of the task object
+ */
+public class TaskType {
+    public static enum taskType {
```
Edwin1022: I like how you use `enum` to represent all the task types.

src/main/java/Milo/TaskList.java
```diff
@@ -0,0 +1,28 @@
...
+
+import Milo.TaskObj.Task;
+import java.util.ArrayList;
+
+/*
+* Represents task list object
+ */
+public class TaskList {
+
+    private final ArrayList<Task> todoList;
```
Edwin1022: Would it be better to just name your ArrayList<Task> attribute here as taskList because there is a task type named Todos but taskList stores other tasks like Deadline tasks and Event tasks as well?

src/main/java/Milo/Ui.java
```diff
@@ -0,0 +1,165 @@
...
+                (˚ˎ 。7 \s
+                 |、˜〵         \s
+                 じしˍ,)ノ
+                """;
+    private final String cat1 = """
+                 ∧,,,∧
+                ( ̳• · •̳)
+                /    づ♡
+                """;
+    private static final String hLine = "____________________________________________________________\n";
```
Edwin1022: This works, but you could also try something like separationLine?

src/main/java/Milo/Parser.java
```diff
@@ -0,0 +1,133 @@
...
+    public static void readInput(String userInput, ArrayList<Task> todoList) {
+        String[] arrOfInput = userInput.split(" ", 2);
+        String action = arrOfInput[0];
+        switch (action) {
+            // Show list
+            case "list":
+                Ui.printList(todoList);
+                break;
+            // Find task
+            case "find":
```
cweijin: I like how you use switch statement on the action keyword, but do check that your switch statement follows the coding convention (no indentation for case clauses).

src/main/java/Milo/Parser.java
```diff
@@ -0,0 +1,133 @@
...
+            // Event
+            case "event":
+                // Check case where event empty
+                if (arrOfInput.length == 1) {
+                    Ui.printError(TaskType.taskType.EVENT, "The description of an event cannot be empty");
+                } else {
+                    // Check case where event command is not properly formatted
+                    String[] eventDesc = arrOfInput[1].split("/from | /to", 3);
+                    if (eventDesc.length != 3) {
+                        Ui.printError(TaskType.taskType.DEADLINE, "Invalid event command\n Proper formatting: deadline <task description> + /from + <starting date description> + /to + <ending date description");
```
cweijin: Perhaps can use line wrapping? (Line length should be no longer than 120 chars)

src/main/java/Milo/Ui.java
```diff
@@ -0,0 +1,165 @@
...
+                String invalidDateMessage = oops + "date is not properly formatted\n " +
+                        "It should be formatted as follow: YYYY-MM-DD\n";
+                System.out.println(hLine + invalidDateMessage + hLine);
+                break;
+            default:
+                String defaultMessage = oops + "I'm sorry, I don't know what that means ;-;\n";
+                System.out.println(hLine + defaultMessage + hLine);
+        }
+    }
+}
```
cweijin: I like how you have a method to handle error printing.

src/main/java/Milo/TaskObj/TaskType.java
```diff
@@ -0,0 +1,15 @@
+package Milo.TaskObj;
+
+/*
+ * Represents the different possible task type of the task object
+ */
+public class TaskType {
```
cweijin: Is there any reason why you wrap the `enum` inside the `TaskType` class instead of having just `public enum TaskType`?

src/main/java/Milo/TaskObj/TaskType.java
```diff
@@ -0,0 +1,15 @@
...
+    public static enum taskType {
+        TODO,
+        EVENT,
+        DEADLINE,
+        INVALID,
+
+        DATE
```
cweijin: Reading through your code it seems like these are used to tag error to be handled by `printError` in Ui. Perhaps making it `ErrorType` instead of `TaskType`?

src/main/java/Milo/Ui.java
```diff
@@ -0,0 +1,165 @@
...
+                  ╱|、
+                (˚ˎ 。7 \s
+                 |、˜〵         \s
+                 じしˍ,)ノ
+                """;
+    private final String cat1 = """
+                 ∧,,,∧
+                ( ̳• · •̳)
+                /    づ♡
+                """;
```
cweijin: So cute!! 😻

cweijin (commented):
LGTM! 👍 
Just some coding convention to follow.

PR #445, opened 2024-08-29
[tanwartushar] iP

tanwartushar:
# Donna

> "I'm Donna, [I know everything](https://youtu.be/Ehxl807mUBk?t=291)" 

Donna is here to help you streamline your task management with ease :)

### What Donna Can Do 

- Add  `ToDo`, `Deadline`, or `Event` tasks
- Delete tasks
- List all the tasks
- Mark tasks as complete (or incomplete)
- Save tasks locally (and pick up from where you left!)
<sub>what can she not do?</sub>

**To get Donna's assistance with anything, follow these steps:** 

1. Start the chatbot
2. Enter the command
3. Let Donna handle the rest! :wink:

Features to be added:

- [x] Add a basic GUI
- [ ] Tweak the UI 

##### PS: This is how Donna marks tasks

```
Task task = tasks.markTask(taskIdx, false);
ui.printTaskMarkedMessage(task, false);
```


src/main/java/donna/Donna.java
```diff
@@ -0,0 +1,180 @@
...
+     */
+    private void run() {
+        boolean dataWasLoaded = !tasks.isEmpty();
+        ui.printGreeting(dataWasLoaded);
+
+        Scanner sc = new Scanner(System.in);
+        while (true) {
+            String input = sc.nextLine();
+            try {
+                ParsedCommand result = parser.parse(input);
```
rachelyeohm, line 56: If you don't need the parser class to be unique to each instance, it could be better to make it static as it's for utility.

src/main/java/donna/Donna.java
```diff
@@ -0,0 +1,180 @@
...
+                case "exit":
+                    ui.printGoodbyeMessage();
+                    storage.saveTasks(tasks);
+                    sc.close();
+                    return;
+                case "list":
+                    ui.printTaskList(tasks);
+                    break;
+                case "mark":
+                    handleMark(result.getArgument1());
```
rachelyeohm: I really like that you separate your command into arguments and command name in the ParsedCommand - it's very neat.

src/main/java/donna/Donna.java
```diff
@@ -0,0 +1,180 @@
...
+        }
+    }
+
+    /**
+     * Handles the marking of a task as not done.
+     *
+     * @param taskNum Index of the task to unmark (index from 1).
+     * @throws DonnaException If the argument is not a valid task number.
+     */
+    private void handleUnmark(String taskNum) throws DonnaException {
```
rachelyeohm: Since handleUnmark, handleAdd etc are all the same type of methods, it might be better to put them in a separate class of their own, or have different types of Commands and ways of executing them.

src/main/java/donna/Storage.java
```diff
@@ -0,0 +1,107 @@
...
+            if (!Files.exists(filePath)) { //ensure that the directory exists- if not, it will be created
+                return tasks; //empty arrayList is returned when no file exists
+            }
+
+            File file = new File(String.valueOf(filePath));
+            Scanner sc = new Scanner(file);
+
+            while (sc.hasNext()) {
+                String line = sc.nextLine();
+                String[] inWords = line.split(" \\| ");
```
rachelyeohm, line 84: I think this could be named more meaningfully / clearly :(

src/main/java/donna/Ui.java
```diff
@@ -0,0 +1,152 @@
...
+    static private void printDashedLine() {
+        System.out.println("____________________________________________________________________");
+    }
+    static private void printDonnaLogo() {
+        System.out.println(" ____      ");
+        System.out.println("|  _ \\  ___  _ __  _ __   __ _ ");
+        System.out.println("| | | |/ _ \\| '_ \\| '_ \\ / _` |");
+        System.out.println("| |_| | (_) | | | | | | | (_| |");
+        System.out.println("|____/ \\___/|_| |_|_| |_|\\__,_|");
+        System.out.println();
```
rachelyeohm: so cute haha

src/main/java/donna/Ui.java
```diff
@@ -0,0 +1,152 @@
...
+            System.out.println("No tasks added to the list yet."+ "\n"
+                    + "use todo / deadline / event to add tasks to the list!");
+
+        } else {
+            System.out.println("There are " + tasks.size() + " task(s) in the list: ");
+            for (int i = 0; i < tasks.size(); i++) {
+                System.out.println((i + 1) + ". " + tasks.get(i));
+            }
+        }
+        printDashedLine();
```
rachelyeohm: I would suggest, if you have to keep writing printDashedLine() before and after each paragraph of text, it would be better to have a special method that does it for you (to avoid repetition)

src/test/java/donna/task/TaskTest.java
```diff
@@ -0,0 +1,38 @@
+package donna.task;
+
+import donna.DonnaException;
+import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.*;
+
+public class TaskTest {
+
+    @Test
+    void testTaskInitialisationWithValidDescription() throws DonnaException {
```
rachelyeohm, line 15: If you're feeling fancy, you could split the test method names into three parts like in the tutorial

src/main/java/donna/Donna.java
tanwartushar: Makes sense. I'll work on it

src/main/java/donna/Ui.java
tanwartushar: Thank you for pointing that out. Fixed it :)

src/main/java/donna/Donna.java
tanwartushar: Thank you!

src/main/java/donna/Donna.java
tanwartushar: That's a valid point. I'll do so.

src/main/java/donna/Storage.java
tanwartushar: True! Will do.

src/test/java/donna/task/TaskTest.java
tanwartushar: Noted xD

src/main/java/donna/Donna.java
```diff
@@ -0,0 +1,228 @@
+package donna;
+
+import donna.parse.ParsedCommand;
+import donna.parse.Parser;
+import donna.task.*;
```
ty-3401: I think the tasks import should be listed explicitly.
https://se-education.org/guides/conventions/java/intermediate.html#statements

src/main/java/donna/DonnaException.java
```diff
@@ -0,0 +1,34 @@
...
+                + "Please use a valid task type \n"
+                + "(todo / deadline / event)\n"
+                + "or you may type list , delete or mark / unmark\n"
+                + "to view, delete or mark tasks");
+    }
+
+    public static DonnaException invalidTaskNumber() {
+        return new DonnaException("No task assigned to this number yet.\n"
+                + "Retry with a valid task number!");
+    }
```
ty-3401: I like how descriptive each of the error messages are! Kudos to you!

src/main/java/donna/Storage.java
```diff
@@ -0,0 +1,104 @@
+package donna;
+
+import donna.task.*;
```
ty-3401: Same as a previous comment. Imports should be listed explicitly.
https://se-education.org/guides/conventions/java/intermediate.html#statements

src/main/java/donna/Ui.java
```diff
@@ -0,0 +1,173 @@
...
+
+
+    /**
+     * Returns a greeting message to the user.
+     *
+     * @return The greeting message.
+     */
+    public String getGreeting() {
+        StringBuilder greeting = new StringBuilder();
+//        greeting.append(donnaLogo());
```
ty-3401: I think if you're not gonna use this you can delete it. Otherwise I think it should be in line with the rest of the code.
https://se-education.org/guides/conventions/java/intermediate.html#comments

src/main/java/donna/Ui.java
```diff
@@ -0,0 +1,173 @@
...
+    static private String dashedLine() {
+        return "____________________________________________________________________\n";
+    }
+
+    static private String donnaLogo() {
+        return " __      \n"
+                + "|  _  \\  __  _ __   _ _   __  \n"
+                + "|  |  |  | / _ \\| '_  \\ | '_ \\ / _` |\n"
+                + "|  |_| |  (_) |  | |  |  |  |   | (_|  |\n"
+                + "|___/ \\__/|_| |_|_|  |_|\\_,_|\n";
```
ty-3401, line 28: I love the ASCII art. I've also done the same in my code.

ty-3401 (commented):
Sorry for the late review! I hope you find them useful. I don't believe there is much I can find wrong with your code other than some nits.

src/main/java/donna/Donna.java
tanwartushar: yes 😅 will fix it.

src/main/java/donna/Storage.java
tanwartushar: will fix it.

src/main/java/donna/Ui.java
tanwartushar: You're right. I left it for the text UI, but I can remove it now.

tanwartushar:
> Sorry for the late review! I hope you find them useful. I don't believe there is much I can find wrong with your code other than some nits.

Thank you for the review! Will do the needful.

PR #462, opened 2024-08-29
[Nasya Myrilla Wijaya] iP

myrilla12:
# Eevee Chatbot
*Named after my dog* 🦮 
> This is my original quote

- easy to use
- efficient user interface

All you need to do is:
1. **Download** it
2. Open it
3. Type in commands
4. Make a task list ~~yay~~

```java
public class Main {
    }
}
```
This is not actually my `main` class. 

Things to do: 
- [ ] Make a GUI
- [ ] Reminders

I learned how to do this from [Github documentation](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#links).


src/main/java/eevee/Eevee.java
```diff
@@ -0,0 +1,161 @@
...
+    public void run() {
+        ui.printGreeting();
+
+        while (true) {
+            String input = ui.getInput();
+
+            try {
+                Parser.Command command = parser.parseCommand(input);
+
+                switch (command) {
```
InfinityTwo: Using t, s, d, e as names in the entire switch statement can be a little troublesome to remember since there is an increasing number of single lettered variables. Would writing it in full be better?

src/main/java/eevee/Event.java
```diff
@@ -0,0 +1,56 @@
...
+     */
+    public Event(String description, String from, String to) {
+        super(description.trim());
+
+        // Handle date given in format yyyy-mm-dd
+        String startDate = from.trim();
+        String endDate = to.trim();
+        try {
+            LocalDate date = LocalDate.parse(startDate);
+            LocalDate date2 = LocalDate.parse(endDate);
```
InfinityTwo: Would it better to write endDateParsed instead of date2? same with date as well

src/main/java/eevee/Ui.java
```diff
@@ -0,0 +1,59 @@
+package eevee;
+
+import java.util.Scanner;
+
+/**
+ * Represents the CLI UI for the application.
+ */
+public class Ui {
+    private final String divider = "____________________________________________________________\n";
```
InfinityTwo: Shouldn't final var names be in SCREAMING_SNAKE_CASE?

InfinityTwo (commented):
I see a lot of single character variable names, which may be a bit annoying to follow but those variables are rather straightforward so it's hard to say if it's better to change it. Other than that, the rest LGTM.

warg-void (commented):
LGTM!

src/main/java/eevee/Eevee.java
warg-void, line 1: I like the use of try catch blocks. In the main run loop could the try catch block be split to cover just the input command?

src/main/java/eevee/Deadline.java
warg-void, line 1: Could by in the Deadline constructor be given a different name as it is unformatted while the by in the class is formatted?

PR #477, opened 2024-08-29
[skyl3-r] iP

skyl3-r:
# Barcus
> "Before you eat the elephant, make sure you know what parts you want to eat." - Todd Stocker [(source)](https://www.lifestarr.com/blog/11-to-do-list-quotes-to-help-you-begin-tackling-things-today)

Barcus keeps track of what tasks you want to do (or parts of an elephant you want to eat 🐘 😋 ). It's
- text-based
- easy to use
- ~~silly goofy~~ very serious

All you need to do is,
1. download it from [here](https://github.com/skyl3-r/ip/releases/tag/v0.1)
2. run it
3. add in your tasks
4. let barcus fix your life 👍 

> [!NOTE]
> Ensure you have **Java 17** installed

Features:
- [x] Add tasks
- [x] Delete tasks
- [ ] GUI (coming soon)

```Java
public class Main {
    }
}
```


src/main/java/barcus/command/MarkCommand.java
```diff
@@ -0,0 +1,36 @@
...
+     * @param pos index of item to mark
+     */
+    public MarkCommand(int pos) {
+        this.pos = pos;
+    }
+
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        if (pos > 0 && pos <= tasks.getLength()) {
+//                        tasks.get(pos - 1).markDone();
```
valhrd: Is this still necessary?

src/main/java/barcus/command/UnmarkCommand.java
```diff
@@ -0,0 +1,36 @@
...
+     * @param pos index of item to unmark
+     */
+    public UnmarkCommand(int pos) {
+        this.pos = pos;
+    }
+
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        if (pos > 0 && pos <= tasks.getLength()) {
+//                        tasks.get(pos - 1).unmarkDone();
```
valhrd: Is this commented out code still necessary?

src/main/java/barcus/task/Event.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Task with from and to dates
+ */
+public class Event extends Task {
+    protected LocalDateTime from;
+    protected LocalDateTime to;
+
+    private final static DateTimeFormatter fromFormatter = DateTimeFormatter.ofPattern("dd/MM/yyyy HH:mm");
+    private final static DateTimeFormatter toFormatter = DateTimeFormatter.ofPattern("MMM dd yyyy HH:mm");
```
valhrd: This code looks to be repeated in the Deadline class as well, could we try to "shift" it up to its parent class and handle formatting there instead?

src/main/java/barcus/task/Task.java
```diff
@@ -0,0 +1,77 @@
...
+        String status = this.isDone ? "1" : "0";
+        return status + " | " + this.description;
+    }
+
+    /**
+     * Returns whether the description contains the substring
+     * @param toFind string to find
+     * @return true if description contains the substring
+     */
+    public boolean containsSubstring(String toFind) {
```
valhrd, line 78: Just a minor preference of changing this method name to "find" but otherwise there's no issue here.

src/main/java/barcus/tasklist/TaskList.java
```diff
@@ -0,0 +1,119 @@
...
+import barcus.task.Task;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Class to encapsulate the list of tasks
+ */
+public class TaskList {
+    private ArrayList<Task> tasks;
+    private int length;
```
valhrd, line 14: Is the private field "length" necessary since we can call this.tasks.size() in its place?

src/main/java/barcus/tasklist/TaskList.java
```diff
@@ -0,0 +1,119 @@
...
+
+import barcus.task.Task;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Class to encapsulate the list of tasks
+ */
+public class TaskList {
+    private ArrayList<Task> tasks;
```
valhrd, line 13: Just a minor preference but maybe rename to taskList or listOfTasks to be a bit more descriptive, otherwise no issues at all.

src/main/java/barcus/command/Command.java
```diff
@@ -0,0 +1,14 @@
...
+import barcus.ui.Ui;
+import barcus.storage.Storage;
+import barcus.exception.BarcusException;
+
+/**
+ * Abstract command class
+ */
+public abstract class Command {
+    public abstract void execute(TaskList tasks, Ui ui, Storage storage) throws BarcusException;
+    public abstract boolean isExit();
```
valhrd, line 30: Many commands override this method to return false except for the ExitCommand class. Is it possible to make this method in the Command class concrete and override it to return true only for the ExitCommand class?
```suggestion
    public boolean isExit() {
        return false;
    }
```

valhrd (commented):
Nice implementation of the Command class! Only some minor preferences when it comes to naming some fields and simplifying some stuff to avoid repeated code blocks, but overall very well-structured code.

src/main/java/barcus/parser/Parser.java
```diff
@@ -0,0 +1,123 @@
...
+        } else if (words[0].equals("find")) {
+            if (words.length == 1) {
+                throw new BarcusException("please include what word(s) you want to find after 'find'");
+            } else {
+                return new FindCommand(String.join(" ", Arrays.copyOfRange(words, 1, words.length)));
+            }
+        }
+
+        return new UnknownCommand();
+    }
```
EyuGongYi: Could use enum and switch statements to make the code look cleaner? It looks abit harder to read for now.
Do you think you can switch up the logic for checking if the command is valid? for example taking out the words.length check to check for multiple commands at once instead of 1 by 1

src/main/java/barcus/task/Deadline.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+/**
+ * Task with a date to do by
+ */
+public class Deadline extends Task {
+    private static final DateTimeFormatter fromFormatter = DateTimeFormatter.ofPattern("dd/MM/yyyy HH:mm");
```
EyuGongYi, line 11: Maybe Deadline and Event can extend of a TimedTask? so that there is no need for multiple formatters.

src/main/java/barcus/ui/Ui.java
```diff
@@ -0,0 +1,59 @@
...
+     */
+    public void talk(String s) {
+        System.out.println("Barcus: " + s);
+    }
+
+    /**
+     * Prints welcome message
+     */
+    public void showWelcome() {
+        talk("Beep bop! Hello I am Barcus, ready to be of assistance!\n"
```
EyuGongYi: Like how you stored all the constant information into UI and appending to the string as needed depending on the command.

src/main/java/barcus/storage/Storage.java
```diff
@@ -0,0 +1,104 @@
...
+                            lineSplit[2],
+                            lineSplit[1].equals("1"),
+                            lineSplit[3]
+                    ));
+                } else if (lineSplit[0].equals("E")) {
+                    tasks.addTask(new Event(
+                            lineSplit[2],
+                            lineSplit[1].equals("1"),
+                            lineSplit[3],
+                            lineSplit[4]
```
EyuGongYi: Could separate this into a separate Parser class? Can help keep the code clean if you plan to store more types of data other than the current 3

EyuGongYi (commented):
Good job overall, the code standard is great and i cant really find any major things, just small things to nitpick about

PR #493, opened 2024-08-29
[Shane] iP

Shantaa:
```c
int main() {
    printf("Hello, Joseph!");
    return 0;
}
```

# Joseph helps you keep track of [stuff](https://www.stuffd.com/sg/). You can

- Add/Delete tasks
- Mark/unmark tasks
- Work with different types of tasks with different details

Just set it up following these steps:

1. Download it from [~~here~~](https://github.com/Shantaa/ip/releases)
2. Run it
3. Add your tasks
4. And you're set! :smile:

Features

- [x] Managing Tasks
- [ ] Managing Deadlines
- [ ] Reminders

> Thanks!

`inline code here`


k-hian (commented):
LGTM. No obvious naming issues. Good work.

src/main/java/Joseph/UI.java
```diff
@@ -0,0 +1,180 @@
...
+
+    public UI() {
+        this.scanner = new Scanner(System.in);
+    }
+
+    /**
+     * Prints a line separator.
+     */
+    public void printLine() {
+        String LINE = "----------------------------------";
```
NicholasCWK: I think variable names with all uppercase should be reserved for constant variables. Perhaps you could consider using a static final variable? 😅

src/main/java/Joseph/Tasks/Task.java
```diff
@@ -0,0 +1,25 @@
...
+    }
+    public String getDone() {
+        return isDone ? "X" : " ";
+    }
+    public String getDesc() {
+        return this.desc;
+    }
+    public String getDetails() {
+        return this.desc;
+    }
```
NicholasCWK, line 26: It seems that these 2 methods do the same thing, perhaps you could remove the getDetails method?

src/main/java/Joseph/Storage.java
```diff
@@ -0,0 +1,131 @@
...
+                        break;
+                    case "E":
+                        String[] eventDesc = desc.split(" /");
+                        JEvent event = new JEvent(eventDesc[0], eventDesc[1], eventDesc[2]);
+                        if (isDone) {
+                            event.setDone();
+                        }
+                        list.add(event);
+                        break;
+                }
```
NicholasCWK: I think "case" is supposed to be the same indentation as "switch" according to the java coding standard

src/main/java/Joseph/Tasks/TaskList.java
```diff
@@ -0,0 +1,165 @@
...
+     *
+     * Unmarks a task as not done.
+     * @param index The index of the task to be unmarked. User input is 1-indexed.
+     * @throws IndexOutOfBoundsException If no task exists at the specified index.
+     */
+    public void unmarkTask(int index) throws IndexOutOfBoundsException {
+        tasks.get(index - 1).unDone();
+    }
+
+    public int size() {
```
NicholasCWK, line 78: Names representing methods should be verbs according to the java coding standard. Perhaps you could change it to getSize? 😅

NicholasCWK (changes requested):
Overall I found your code to be clear and well structured. I found your code standard very consistent, just take note of the naming of methods and variables.

Also, good job on the detailed JavaDoc and the implementation of the various exceptions!

src/main/java/Joseph/Tasks/Task.java
Shantaa: oh thanks i missed that

Shantaa: I actually meant for getDesc to return just the description, while getDetails would return the description along with the due for Deadlines and the timings for Events (I override getDetails in the Deadline and Event classes), but I can see how it can be misleading. Thanks!

PR #494, opened 2024-08-29
[kevin-pek] iP

kevin-pek:
# Skibidi Chatbot

Skibidi chatbot is a **free** chatbot to help you remember the things you need to do in case your brain rots.

- text-based
- easy to learn
- ~fast~ 🔥🔥🔥 *BLAZINGLY FAST*

To try out this app, follow these steps:

1. download it from [here](https://www.youtube.com/watch?v=dQw4w9WgXcQ)
2. double-click it
3. add your tasks

>"deep sounding quote" ~ important person name, ([for more quotes](https://inspirobot.me/))

## Roadmap

- [x] Managing tasks
- [ ] Managing deadlines
- [ ] Reminders

## Code snippets

You might encounter a significant amount of frustration when using the app since it is currently 100% text based which could lead to suboptimal user experience. To worsen this experience, create and run a file names `Main.java` with the following code:

```java
import java.io.IOException;

public class Main {
	public static void main(String[] args) throws IOException {
		Runtime.getRuntime().exec(new String[]{"java", "-cp", System.getProperty("java.class.path"), "Main"});
	}
}
```


src/main/java/skibidi/Skibidi.java
```diff
@@ -0,0 +1,71 @@
+package skibidi;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+
+import gui.MainGui;
+import javafx.application.Application;
+
+/**
```
somethingfishyfishy: (not a violation)
Just pointing out for single liners, you could write the javadoc comment as a single line.

Eg.
/** Number of connections to this database */
private int connectionCount;

src/main/java/skibidi/Storage.java
```diff
@@ -0,0 +1,73 @@
...
+                FileWriter writer = new FileWriter(file, false);
+                for (int i = 1; i <= taskList.size(); i++) {
+                    AbstractTask task = taskList.getTask(i);
+                    writer.write(task.serialize() + '\n');
+                }
+                writer.close();
+            } else {
+                System.out.println("Failed to save changes to file.");
+            }
+        } catch (TaskList.TaskNotFoundException err) {
```
somethingfishyfishy: can apply polymorphism.

Eg.
catch (Exception e) {System.out.println(e.getMessage())}

src/main/java/skibidi/Skibidi.java
kevin-pek: Thanks for the input! This will help make the codebase more concise and readable.

src/main/java/skibidi/command/AbstractCommand.java
LimKaiWei, line 1: interesting choice of class name, many just use `Command`

src/main/java/skibidi/Storage.java
```diff
@@ -0,0 +1,73 @@
...
+    private String dataPath;
+
+    public Storage(String dataPath) {
+        this.dataPath = dataPath;
+    }
+
+    /**
+     * Save given task list to storage data path. Overrides any existing data on the file.
+     */
+    public void saveTasksToDisk(TaskList taskList) {
```
LimKaiWei, line 28: i like this descriptive method name, here and throughout your code

src/main/java/skibidi/command/AbstractCommand.java
kevin-pek: Right, I added this prefix for abstract classes so it would be clearer at a glance that this was an abstract class. Though I see why people would find this unnecessary too. Going to leave it as it is since it doesn't go against any of the coding standards (to my knowledge).

src/main/java/skibidi/Storage.java
kevin-pek: Thanks!

src/main/java/skibidi/Storage.java
kevin-pek: Good call, a generic Exception handler would suffice for cases where special handling is not needed.

PR #510, opened 2024-08-29
[stephaneleong] iP

stephaneleong:
# Easton :writing_hand:

> “Create a short list of things you need to get done today.” - Matt D'Avella ([My Simple Productivity System](https://www.youtube.com/watch?v=BtiQvhQF8IA))

## Task Management Made Easy (Capabilities
+ _Create_ New Tasks
+ _Delete_ A Task
+ Keep _Track_ Of Your Tasks
+ _Search_ Your Tasks

## Different Kind Of Tasks
1. **Todos**
1. **Deadlines**
1. **Events**

## A TODO List For You
- [x] Download Easton
- [x] Create New Tasks
- [x] Become More Productive

If you want to use Easton in your own project:

```java
    new Easton("file.csv").run();
}
```


src/main/java/easton/Easton.java
```diff
@@ -0,0 +1,263 @@
...
+import easton.model.Deadline;
+import easton.model.Event;
+import easton.model.Task;
+import easton.model.ToDo;
+import easton.view.Ui;
+
+import java.io.*;
+import java.util.ArrayList;
+
+public class Easton {
```
weijianwong, line 22: Perhaps you could include javadoc header comments for classes and methods so that readers can better understand your code.

src/main/java/easton/Easton.java
```diff
@@ -0,0 +1,263 @@
...
+            Ui.divider();
+
+            try {
+                action = getActionFromInput(userInput);
+            } catch (IllegalActionException e) {
+                Ui.displayText(e.getMessage());
+                action = Action.INVALID;
+            }
+
+            switch (action) {
```
weijianwong: Great job with the indentation of the switch statement. Many students often overlook this part so good job for avoiding this mistake!

src/main/java/easton/Easton.java
```diff
@@ -0,0 +1,263 @@
...
+            Ui.divider();
+        }
+
+    }
+
+    public static void main(String[] args) {
+        new Easton("task.csv").run();
+    }
+
+    private void changeTaskStatus(String input, boolean isDone, String message) {
```
weijianwong: Maybe you could move this logic to the Task class?

src/main/java/easton/DateTimeFormatException.java
```diff
@@ -0,0 +1,7 @@
+package easton;
+
+public class DateTimeFormatException extends Exception {
```
weijianwong: Perhaps this could be in another package such as the exceptions package?

src/main/java/easton/Storage.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        if (Files.notExists(filePath)) {
+            Files.createFile(filePath);
+        }
+
+        this.filePath = filePath;
+    }
+
+    public ArrayList<String> retrieve() throws IOException {
```
weijianwong: Perhaps you could consider changing the naming of the functions in this class to make it clearer? You could name this function "retrieveRecord" instead of just "retrieve"?

src/main/java/easton/view/Ui.java
```diff
@@ -0,0 +1,60 @@
+package easton.view;
+
+import java.util.ArrayList;
+import java.util.Scanner;
+
+public class Ui<T> {
```
weijianwong: Why does this class have to be a generic type?

src/main/java/easton/Easton.java
```diff
@@ -0,0 +1,263 @@
+package easton;
```
weijianwong, line 1: Package naming is in lower case. Class names are in PascalCase and variable names are in camelCase. Great job for following the naming conventions.

weijianwong (commented):
Overall great job for following the coding style conventions with the correct naming conventions and indentations. One improvement that i could suggest is to write header comments for all your classes and methods to make it clearer for the readers. But overall great job!

src/main/java/easton/Easton.java
stephaneleong: Hi @weijianwong! yes I was late in pushing the javadoc comments, I have included it for all my public classes and methods.

src/main/java/easton/Easton.java
stephaneleong: Okay I will look into this!

src/main/java/easton/DateTimeFormatException.java
stephaneleong: okay. I will move all the exception into an exception package

src/main/java/easton/Storage.java
stephaneleong: Made the changes! thank you

src/main/java/easton/view/Ui.java
stephaneleong: Well I agree it is not necessary because of the scale of this ip, however, when I was planning out the design I thought it would also make sense to make this class generic so that if there were other Classes we would want to display (e.g. listing) I would not need to write code. 

However, I would appreciate your input on this thought of my @weijianwong. Thank you!

src/main/java/easton/Easton.java
stephaneleong: @weijianwong looking into the logic of this function, changeTaskStatus(). I'm worried if I shifted the logic into the Task class, specfically the printing of statement, that I would create a dependency assiocation between Task and the Ui classes which I forsee the code becoming more buggy if one of the parts were to break. But, yes input is appreciated.

src/main/java/easton/Easton.java
```diff
@@ -0,0 +1,396 @@
...
+    }
+
+    /**
+     * Changes the status of a given task.
+     *
+     * @param input Input from the prompt.
+     * @param isDone Is the task done.
+     * @param message Message displayed to the user interface.
+     */
+    public void changeTaskStatus(String input, boolean isDone, String message) {
```
hazzle23: I like how you use clear descriptive verbs for all of the methods in this class. Good job on following the naming convention!

src/main/java/easton/Action.java
```diff
@@ -0,0 +1,8 @@
+package easton;
+
+/**
+ * Represents that the chatbot can handle.
```
hazzle23: I think the JavaDoc could be more descriptive to better explain what `Action` represents. A possible revision could be 
`*Represents the different types of actions/commands that the chatbot can handle.`

src/main/java/easton/model/Event.java
```diff
@@ -0,0 +1,48 @@
...
+
+import easton.exception.DateTimeFormatException;
+
+/**
+ * Represents an event task.
+ */
+public class Event extends Task {
+
+    private LocalDateTime from;
+    private LocalDateTime to;
```
hazzle23: While variable name `to` is functional, I think you can consider renaming it to something more descriptive for better clarity, like `endDate` or `toDate`.

src/test/java/easton/EastonTest.java
```diff
@@ -0,0 +1,60 @@
+package easton;
+
+import easton.exception.DateTimeFormatException;
+import easton.exception.EmptyDescriptionException;
+import easton.exception.InvalidFormatException;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
```
hazzle23: Based on [this guide](https://se-education.org/guides/tutorials/intellijCodeStyle.html#tweak-import-order), I think static imports should be ordered first.

src/test/java/easton/EastonTest.java
```diff
@@ -0,0 +1,60 @@
...
+    @Test
+    public void createDeadline_withoutByLabel_exceptionThrown() {
+        String input = " deadline read book";
+        assertThrows(InvalidFormatException.class, () -> {
+            Easton.createDeadline(input);
+        });
+    }
+
+    @Test
+    public void createDeadline_incorrectDateTimeFormat_exceptionThrown() {
```
hazzle23, line 37: Good job on following the test methods naming convention!

hazzle23 (commented):
Overall looks good to me!

PR #534, opened 2024-08-29
[PrishaVP] iP

PrishaVP:
# MeowMeow Chatbot 🐱

> "MeowMeow helps you manage your tasks efficiently, so you can focus on more important things... like chasing laser pointers!" – _Anonymous Cat_

MeowMeow is:

- text-based
- user-friendly
- ~~FAST~~ *SUPER* FAST to use!

All you need to do is,

1. download it from [here](https://github.com/PrishaVP/ip).
2. double-click it.
3. add your tasks.

And it is **FREE!**

Features:

- [x] Task management
- [x] Deadline tracking
- [x] Event scheduling

If you are a Java programmer, you can use it to practice Java too. Here’s the `main` method:

```java
public class Launcher {
    }
}
```


src/main/java/meowmeow.java
```diff
@@ -0,0 +1,94 @@
+import java.util.ArrayList;
+import java.util.Scanner;
+
+public class meowmeow {
```
bbbbcccd: Will it be better for the class name to start with a captial letter?

bbbbcccd (commented):
Lgtm!

src/main/java/meowmeow.java
```diff
@@ -1,3 +1,4 @@
+import java.util.LinkedList;
 import java.util.Scanner;
 public class meowmeow {
```
HanB1n: Should class names be renamed using CamelCase?

src/main/java/meowmeow.java
```diff
@@ -46,24 +46,32 @@ public static void main(String[] args) {
...
+                    String by = parts[1];
+                    Deadline deadline = new Deadline(description, by);
+                    list.add(deadline);
+                    System.out.println("Got it. I've added this task:");
+                    System.out.println("  " + deadline);
+                    System.out.println("Now you have " + list.size() + " tasks in the list.");
+                }
             } else if (input.startsWith("event ")) {
                 // Add an Event task
                 String[] parts = input.substring(6).split(" /from | /to ");
```
HanB1n: Would it be better if there was error handling in this steps?
Perhaps using try and catch blocks would help with possible Index Errors?

HanB1n (commented):
The code was generally well written and easy to understand.
A general suggestion: Using switch case might be good to improve code readability and reduce the amount of if else statements.

PR #538, opened 2024-08-29
[Joel Tio] iP

joeltio:
# Ip Man

> I wish to fight ten men.
> — Ip Man

Ip Man offloads your mind of daily tasks so that you can be better at [Wing Chun](https://en.wikipedia.org/wiki/Wing_Chun). It's:

- 100% _Gluten-free_ and _organic_ 🌾
- Made only with ✨natural✨ ideas (no artificial LLM code)
- Good for your bones 🦴
- Very demure, very mindful 😇

To become a disciple, you just need to follow the soon-to-be-simplified process of:

1. Clone the repository from [here](https://github.com/joeltio/ip/tree/master)
2. Run the Gradle task `shawdowJar` with your preferred IDE
3. Run the produced Jar file (that can be found in `build/libs/ipMan.jar`) with `java -jar ipMan.jar`
4. Become Bruce Lee

Features:

- [x] Managing tasks (To-dos, Deadlines and Events)
- [x] Storing tasks even after program is closed
- [x] Find tasks by name
- [x] Mark tasks as done
- [ ] Many many more features to be revealed!

If you are a Java programmer and the white font in your dark-themed terminal hurts your eyes, consider overriding the `Ui` interface with your very own #3c4043 on #000000 themed UI!
```java
public interface Ui {
    // Inspired by https://nus-cs2103-ay2425s1.github.io/website/schedule/week3/project.html
    void showWelcome();
    Command readMessage();
    void showMessage(String message);
    void showMessageF(String message, Object... args);
    void showError(String message);
    void showGoodbye();
}
```


src/main/java/ipman/commands/MarkCommand.java
```diff
@@ -0,0 +1,25 @@
+package ipman.commands;
+
+import ipman.models.Task;
+import ipman.models.TaskList;
+
+public class MarkCommand implements Command {
+    private final int taskIndex;
```
cshao02, line 15: Constants should be named in all caps?

src/main/java/ipman/IpMan.java
```diff
@@ -0,0 +1,38 @@
...
+import ipman.models.TasksFileManager;
+import ipman.ui.CliUi;
+import ipman.ui.Ui;
+
+
+public class IpMan {
+    private static final Ui ui = new CliUi();
+    private static final TaskList tasks = new TaskList();
+    private static final TasksFileManager fileManager = new TasksFileManager("./data/ipman.txt");
+    private static final Context context = new Context(tasks, ui, fileManager);
```
cshao02: I noticed the same issue in several places too, but perhaps constants should be named in all caps?

src/main/java/ipman/ui/CliUi.java
```diff
@@ -0,0 +1,48 @@
...
+        return Parser.parseCommand(message);
+    }
+
+    @Override
+    public void showMessage(String message) {
+        System.out.println(message);
+    }
+
+    @Override
+    public void showMessageF(String message, Object... args) {
```
cshao02: Perhaps a more descriptive name can be used here as it is not clear what the letter F means

src/test/java/ipman/commands/DeleteCommandTest.java
```diff
@@ -0,0 +1,38 @@
+package ipman.commands;
+
+import ipman.models.DummyFileManager;
+import ipman.models.TaskList;
+import ipman.models.ToDo;
+import ipman.ui.SpyUi;
+import org.junit.jupiter.api.Test;
```
cshao02: Perhaps similar import statements should be grouped together?

src/test/java/ipman/ui/SpyUi.java
```diff
@@ -0,0 +1,47 @@
+package ipman.ui;
+
+import ipman.commands.Command;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class SpyUi implements Ui {
+    private final List<String> record = new ArrayList<>();
```
cshao02: Perhaps plural form should be used for collection of items

cshao02 (commented):
LGTM. Just a few nits to fix.

src/main/java/ipman/commands/MarkCommand.java
joeltio: Oh, in my opinion, constants which need to have all caps are constants that are inherently constant: things that will never change, things that are basically macros like `#DEFINE` in C. In other cases, I think it is alright to leave it as camel case.

Backing up my point, [the code convention](https://se-education.org/guides/conventions/java/intermediate.html) does not technically specify what constitutes a constant. But the [Google code style](https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names) which the code convention falls back on puts it more clearly:

> Constant names use `UPPER_SNAKE_CASE`: all uppercase letters, with each word separated from the next by a single underscore. But what is a constant, exactly?
>
> Constants are static final fields whose contents are deeply immutable and whose methods have no detectable side effects. Examples include primitives, strings, immutable value classes, and anything set to `null`. If any of the instance's observable state can change, it is not a constant. Merely intending to never mutate the object is not enough.

src/main/java/ipman/IpMan.java
joeltio: I hold the same reasons for keeping these fields camel case as above too

src/main/java/ipman/ui/CliUi.java
joeltio: I was a bit conflicted on how I should name this method. On one hand, the code convention tells me that I need to use camel case. On the other, nobody has ever named a formatting print function `printF`. This method is supposed to be the `printf` of displaying messages.

What do you think?

src/test/java/ipman/commands/DeleteCommandTest.java
joeltio: You're right, I realised my checkstyle did not run on test classes. I have corrected this.

src/test/java/ipman/ui/SpyUi.java
joeltio: Hmm, I think this field is a bit ambiguous. I was thinking of record as a collective noun i.e. I keep one record of all the messages shown. I have renamed this field to `recordedMessages` instead of `records`.

joeltio:
Thanks for taking the time to review my PR @cshao02. I appreciate that you do so even though I have been committing a lot the past few days and doing so on top of your assigned PR.

src/main/java/ipman/IpMan.java
```diff
@@ -0,0 +1,41 @@
+package ipman;
+
+import ipman.commands.Command;
```
marraknas, line 3: I think it's good that you  put the import statements in lexicographical order which is a common style error that people make!

src/main/java/ipman/commands/CreateToDoCommand.java
```diff
@@ -0,0 +1,38 @@
...
+ * @see ToDo
+ * @see Context
+ * @see TaskList
+ */
+public class CreateToDoCommand implements Command {
+    private final String name;
+
+    /**
+     * Creates command that creates a particular to-do
+     * @param name to-do name
```
marraknas, line 22: Should there be a whitespace between the first line of javadocs and the next line? I mean this was repeated everywhere in the code so I was just wondering.

src/main/java/ipman/commands/InvalidIndexException.java
```diff
@@ -0,0 +1,24 @@
...
+public class InvalidIndexException extends IndexOutOfBoundsException {
+    /**
+     * Constructs an <code>InvalidIndexException</code> with the error message
+     * indicating what is the expected range of index and actual given index.
+     * @param capacity size of the list
+     * @param actual requested index of the 1-indexed list
+     */
+    public InvalidIndexException(int capacity, int actual) {
+        super(String.format(
+            "Invalid index for list. Expected a number from 1 to %d, got %d",
```
marraknas, line 20: A lot of people miss this and put the comma in the next line! Good that you did it right 😁

src/main/java/ipman/models/Task.java
```diff
@@ -0,0 +1,46 @@
+package ipman.models;
+
+/**
+ * Represents a task that has a name and may or may not be completed.
+ */
+public abstract class Task {
+    protected boolean isDone = false;
```
marraknas: just curious, how did you decide on these access modifiers?

src/main/java/ipman/ui/CliUi.java
```diff
@@ -0,0 +1,53 @@
...
```
marraknas: Agreed that a better method name could be used in this case!

src/test/java/ipman/parser/ParserTest.java
```diff
@@ -0,0 +1,130 @@
+package ipman.parser;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import java.time.format.DateTimeParseException;
+
+import org.junit.jupiter.api.Test;
+
+public class ParserTest {
```
marraknas, line 10: A good amount of tests! Very comprehensive 🥳

marraknas (commented):
lgtm! Code is well written and tests are very comprehensive. Maybe a few whitespaces and style issues but otherwise really good!

src/main/java/ipman/commands/CreateToDoCommand.java
joeltio: Hmm I think it's not a big deal, but I think it's a good idea to be consistent and have a newline before the [at]s

src/main/java/ipman/models/Task.java
joeltio: In this case, I consider these fields to belong mainly to the classes that inherit this Task class (since Task can't be instantiated anyways). I think that it should be "private" to the subclass i.e. protected.

On the other hand, I could very well just expose `isDone` with a getter and disallow subclasses from modifying it. I'll update this to private and implement getters so that testing is easier. I will be able to check if a Task is marked by some command if I had a getter anyways.

src/main/java/ipman/ui/CliUi.java
cshao02: Perhaps you could use showMessageFormatted or displayMessage, anything that can convey the idea of what the method is doing sufficiently

src/main/java/ipman/ui/CliUi.java
joeltio: Resolved by https://github.com/nus-cs2103-AY2425S1/ip/pull/538/commits/338c4f7e995a96a70ec07341c163301fa9ab3e75

src/main/java/ipman/ui/CliUi.java
joeltio: Resolved by https://github.com/nus-cs2103-AY2425S1/ip/pull/538/commits/338c4f7e995a96a70ec07341c163301fa9ab3e75

src/main/java/ipman/commands/CreateToDoCommand.java
joeltio: Resolved by https://github.com/nus-cs2103-AY2425S1/ip/pull/538/commits/38439567a10a1f9d6c06742e548b97b801ce2548

src/main/java/ipman/models/Task.java
joeltio: Resolved by https://github.com/nus-cs2103-AY2425S1/ip/pull/538/commits/abf860ce968af86f71947ba360d1db9783c63039

PR #550, opened 2024-08-30
[thisisaditya17}] iP

thisisaditya17:
### KillJoy

KillJoy, your tech-savvy personal assistant, brings efficiency and a touch of sass to task management. It's:

- **command line interface**
- **simple to use**
- **SMART & FUN**

Just:

1. **Download it from here.**
2. **Run it.**
3. **Tell KillJoy what you need.**
4. **Sit back and let KillJoy organize your life with a playful banter. 😉**

And yes, it's **FREE!**

#### Features:

- **Task management**
- **Playful interaction**

For those who love coding, KillJoy is a fun way to practice your programming skills. Here's the main method:

```java
public class Main {
        KillJoy kj = new KillJoy();
        kj.start();
    }
}
```

**Let KillJoy take the load off your mind and put a smile on your face!**


src/main/java/killjoy/main/KillJoy.java
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * Decreases the task count by 1.
+     */
+    public void decreaseTaskCount() {
+        this.taskCount--;
+    }
+
+    /**
+     * Adds a task to the task list.
```
SwaminathanViswa, line 52: There should be an empty line between the description and the parameter section. The same goes for the rest of the javadocs :)

src/main/java/killjoy/main/KillJoy.java
```diff
@@ -0,0 +1,134 @@
...
+     * Starts the KillJoy application.
+     */
+    public void start() {
+        System.out.println(ui.getLogoString());
+        System.out.println(ui.getWelcomeString());
+        Scanner user = new Scanner(System.in);
+
+        saveAndLoad.loadTasks();
+
+        while(true) {
```
SwaminathanViswa: there should be a space after while 💭

src/main/java/killjoy/processing/ProcessTasks.java
```diff
@@ -0,0 +1,200 @@
...
+        if (inputSplitBySlash[0].split(" ").length == 1) {
+            ui.printLine();
+            ui.displayInvalidCommandFormatMessage();
+            ui.printLine();
+            return;
+        }
+
+        ui.printLine();
+        switch (taskType) {
+        case TODO: {
```
SwaminathanViswa, line 50: According to the coding standard website [here](https://se-education.org/guides/conventions/java/intermediate.html), there is no need for a pair of curly braces {} after each case. 😄

src/main/java/killjoy/processing/Parser.java
```diff
@@ -0,0 +1,139 @@
...
+        }
+        return inputSplitBySlash[1].replaceFirst("from ", "");
+    }
+
+    /**
+     * Parses the user input to determine the to date of the task.
+     * @param input The user input.
+     * @return The to date of the task.
+     */
+    public static String getTo(String input) {
```
SwaminathanViswa: maybe getToString sounds better than getTo as getTo sounds like its going from one location to another? 🤔

src/main/java/killjoy/processing/Parser.java
```diff
@@ -0,0 +1,139 @@
...
+        }
+        return inputSplitBySlash[1].replaceFirst("by ", "");
+    }
+
+    /**
+     * Parses the user input to determine the from date of the task.
+     * @param input The user input.
+     * @return The from date of the task.
+     */
+    public static String getFrom(String input) {
```
SwaminathanViswa: Similar to getTo, getFrom could be renamed as getFromString or getFromDeadline

src/main/java/killjoy/processing/Parser.java
```diff
@@ -0,0 +1,139 @@
...
+        return inputSplitBySlash[0].replaceFirst("todo ", "")
+                .replaceFirst("deadline ", "").replaceFirst("event ", "");
+    }
+
+    /**
+     * Parses the user input to determine the by date of the task.
+     * @param input The user input.
+     * @return The by date of the task.
+     */
+    public static String getBy(String input) {
```
SwaminathanViswa: Similar to getTo, getBy could be renamed as getByString or getByDeadline

SwaminathanViswa (commented):
Overall it is very good and only a bit of changes to be made to ensure code quality! Well done on the code namings as majority were well named and easy to understand! 😃

src/main/java/killjoy/processing/Parser.java
sumomomomomo: I agree, "getTo" can be ambiguous. However I would suggest renaming it to "getToDate".

src/main/java/killjoy/main/UserInterface.java
```diff
@@ -0,0 +1,109 @@
...
+        printLine();
+        System.out.println("    AGWHAHH!!! What'ya sayin' dawgg??");
+        printLine();
+    }
+
+    public void displayAddedTaskMessage() {
+        System.out.println("    Yo Dawgg!! Added this task:");
+        System.out.println("    " + killJoy.getTask(killJoy.getTaskCount() - 1));
+        if (killJoy.getTaskCount() == 1) {
+            System.out.println("    Now you have " + killJoy.getTaskCount() + " task in the list.");
```
sumomomomomo: You could replace "killjoy.getTaskCount()" with "1".
```suggestion
            System.out.println("    Now you have 1 task in the list.");
```

src/main/java/killjoy/processing/Parser.java
```diff
@@ -0,0 +1,139 @@
...
+        }
+        return new String[]{matcher.group(1), matcher.group(3)};
+    }
+
+    /**
+     * Parses the input to get the date and time.
+     * @param dateTime The date and time.
+     * @return The date and time.
+     */
+    public static LocalDateTime parseDateTime(String dateTime) {
```
sumomomomomo: What "parse" means could be made clearer. Parse here returns a LocalDateTime, but parse below in parseTaskInfo returns a String[]. Maybe consider something like "parseStringToLocalDateTime".
```suggestion
    public static LocalDateTime parseStringToLocalDateTime(String stringDateTime) {
```

src/main/java/killjoy/processing/ProcessTasks.java
```diff
@@ -0,0 +1,200 @@
...
+            ui.printLine();
+        }
+    }
+
+    /**
+     * Creates tasks from the task information that is loaded from the save file.
+     * @param taskInfo The task information.
+     */
+    public void createTasks(String taskInfo) {
+        String[] parts = Parser.parseTaskInfo(taskInfo);
```
sumomomomomo, line 184: I like that you named it as the plural form "parts".

src/main/java/killjoy/task/Deadline.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Represents a task of type Deadline.
+ */
+public class Deadline extends Task {
+    private LocalDateTime by;
+    private static final DateTimeFormatter FORMATTER = DateTimeFormatter.ofPattern("dd MMM yyyy HH:mm:ss");
```
sumomomomomo: I feel "FORMATTER" is ambiguous. Looking at your code, its only purpose is to format the date in toString(). Consider:
```suggestion
    private static final DateTimeFormatter TO_STRING_FORMATTER = DateTimeFormatter.ofPattern("dd MMM yyyy HH:mm:ss");
```
Another small nit but consider moving FORMATTER to Task so you don't redefine it in every subclass of Task.

sumomomomomo (commented):
Generally good. Suggested some changes to make naming less ambiguous.

src/main/java/killjoy/main/UserInterface.java
```diff
@@ -0,0 +1,109 @@
...
+        return EXIT_STRING;
+    }
+
+    public String getWelcomeString() {
+        return WELCOME_STRING;
+    }
+
+    public String getLogoString() {
+        return LOGO_STRING;
+    }
```
Justin-Yeo: i like the clear naming of the methods! 💯

src/main/java/killjoy/main/KillJoy.java
```diff
@@ -0,0 +1,134 @@
...
+import killjoy.task.Deadline;
+import killjoy.task.Event;
+import killjoy.task.Task;
+import killjoy.task.Todo;
+import killjoy.processing.ProcessTasks;
+import killjoy.processing.Storage;
+
+import java.time.LocalDateTime;
+import java.util.ArrayList;
+import java.util.Scanner;
```
Justin-Yeo: if im not mistaken the ordering of imports are wrong, just a slight change! 👍

src/main/java/killjoy/processing/Parser.java
Justin-Yeo: i think the naming of the method could be something clearer to prevent ambiguity!

src/main/java/killjoy/task/Task.java
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * Changes the status of the task.
+     */
+    public void changeStatus() {
+        if (this.isDone) {
+            this.isDone = false;
+        } else {
+            this.isDone = true;
+        }
+    }
```
Justin-Yeo: i like the naming but perhaps could give greater details by naming it as changeTaskStatus instead?

Justin-Yeo (commented):
great work! just minor comments on the naming 💯

src/main/java/killjoy/processing/ProcessTasks.java
thisisaditya17: I asked prof about it. He said that if it is not specified you can do it according your own liking but it has to be consistent. The thing is without braces, the variables declared can be used by other cases as well. To prevent it, I enclosed them in {}.

src/main/java/killjoy/main/KillJoy.java
thisisaditya17: Can you specify what type of ordering should i follow?

src/main/java/killjoy/main/UserInterface.java
thisisaditya17: May I know what would be the purpose of it?
I mean will this make a difference?😅😅

src/main/java/killjoy/task/Task.java
```diff
@@ -0,0 +1,66 @@
...
+
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); // mark done task with X
+    }
+
+    public String getDescription() {
+        return this.description;
+    }
+
+    public boolean getTaskStatus() {
```
shadhankkk, line 60: maybe you can change this boolean function name to something that follows the is, has, was, etc. format

e.g: isTaskActive() or something like that

src/main/java/killjoy/task/Event.java
```diff
@@ -0,0 +1,30 @@
...
+        super(description, TaskType.EVENT);
+        this.from = from;
+        this.to = to;
+    }
+
+    @Override
+    public String getTaskInfo() {
+        return "EVENT|" + String.valueOf(isDone ? 1 : 0) + "|" + this.description + "|" +
+                String.valueOf(this.from).replace("T", " ") + "|"
+                + String.valueOf(this.to).replace("T", " ") + "\n";
```
shadhankkk, line 29: I like your use of line breaks here

shadhankkk (commented):
Great code overall!

PR #552, opened 2024-08-30
[tohhengxing] iP

tohhengxing:
# A heading (Wojak)

- a
- bullet
- list

1. a
2. numbered
3. list

```java
public class Main {
  // a fenced code block (with syntax highlighting)
    }
}
```

- [ ] a
- [ ] task
- [ ] list

> a block quote

[a hyper link](https://www.google.com)

an emoji 👎 

`inline.code()`

some text formatting: **bold**, _italic_, ~strikethrough~


src/main/java/controllers/InputParser.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    private boolean isDeleteCommand(String command) {
+        String deleteRegex = "delete (100|[1-9]|[1-9][0-9])";
+        Pattern deletePattern = Pattern.compile(deleteRegex);
+        Matcher deleteMatcher = deletePattern.matcher(command);
+        return deleteMatcher.matches();
+    }
+
+    private String extractStringBetweenTwoSubStrings(String command, String prefix, String byMarker) {
```
ckclion, line 230: I like how the method name explains the method

src/main/java/Wojak.java
```diff
@@ -0,0 +1,37 @@
+import java.io.File;
+import java.util.Scanner;
+import controllers.InputParser;
+import controllers.commands.Command;
+import controllers.errors.InvalidCommandError;
+import controllers.errors.InvalidInputError;
+import lib.DbDriverInterface;
+import lib.FileDbDriver;
+import models.*;
```
ckclion: Perhaps it is better to list imports explicitly

src/main/java/lib/FileDbDriver.java
```diff
@@ -0,0 +1,58 @@
...
+import java.io.IOException;
+
+import static java.lang.System.exit;
+
+public class FileDbDriver implements DbDriverInterface {
+
+    public String init() {
+        try {
+            Path currentWorkingDirectory = Paths.get("").toAbsolutePath();
+            Path parentDirectory = currentWorkingDirectory.getParent();
```
ckclion: I like that there is good distinction between the directories

ckclion (commented):
LGTM. Just some nits to fix.

src/main/java/Wojak.java
tohhengxing: Listing imports explicitly is useful for context. Importing using wildcard is useful for not writing more code than necessary. I find that only if the project is bigger you'd want to list all your dependencies.

Even then, most IDEs have code completion for Java.

src/main/java/controllers/InputParser.java
```diff
@@ -0,0 +1,218 @@
...
+                Event event = new Event(description, from, to);
+                return new AddEventCommand(event);
+            } catch (Exception e) {
+                throw new InvalidInputError("Invalid input for event!");
+            }
+        } else {
+            String errorString = String.format("%s doesn't exist as a command", cmd);
+            throw new InvalidCommandError(errorString);
+        }
+    }
```
ritulkrsingh, line 132: Maybe a switch statement can be used here? (the `equals` methods have to be implemented anyways)

src/main/java/controllers/commands/AddDeadlineCommand.java
ritulkrsingh, line 1: I like how there are classes for operation of commands too!

PR #556, opened 2024-08-30
[Harith Bin Nurhisham] iP

harithhh06:
### LunaBot

> You can only manage time if you track it right. - Spica's team [(source)](https://www.spica.com/blog/the-best-time-management-quotes)

Why remember tasks when a bot can do it for you?

Bogged down with too many submissions, events and things to do? Use LunaBot to manage your tasks!

LunaBot is:

- _text-based_
- easy to use
- **SUPER FAST**  🚀
- Updated in real time

All you need to do is,

1. download it from [here](https://github.com/Harithhh06/ip/releases).
2. Start the app
3. add your tasks ✅
4. let it manage your tasks for you 📋
5. And it's **FREE**?!?! 🤯

Features:

- [x] Managing tasks
- [ ] Managing deadlines
- [ ] Scheduling events

```java
public class LunaBot {
        String filepath;
        new LunaBot(filepath).run();
    }
}
```


src/main/java/Storage.java
```diff
@@ -0,0 +1,78 @@
+import java.io.*;
```
nigellqx: Perhaps listing out all the imports might be better according to the coding standards?

src/main/java/LunaBot.java
```diff
@@ -0,0 +1,245 @@
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.Scanner;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+public class LunaBot {
```
nigellqx: Perhaps have a Ui and a Parser class to handle different tasks?

src/main/java/LunaBot.java
```diff
@@ -0,0 +1,245 @@
...
+
+        // Create scanner for user input
+        Scanner scanner = new Scanner(System.in);
+        String input;
+
+        // loop scanner to echo user input
+        while (true) {
+            try {
+                input = scanner.nextLine();
+                if (input.equals("bye")) {
```
nigellqx: Perhaps you can use input.equalsIgnoreCase("bye") to handle cases where user might input "BYE"?

nigellqx: minor nitpick - maybe the use of a switch statement is better for readability?

src/main/java/LunaBot.java
```diff
@@ -0,0 +1,245 @@
...
+            }
+        }
+        System.out.println("___________________________________________________________________");
+    }
+
+    // Method to mark tasks
+    private static void markTask(String input) throws LunaBotException {
+        try {
+            String taskNumString = input.substring(5);
+            // Checks to see if user provided a number
```
nigellqx: I like how you add comments in the nested code to allow readers to know what you are doing at each stage of the code

src/main/java/Storage.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.Date;
+
+public class Storage {
+    private String filePath;
+
+    public Storage(String filePath) {
+        this.filePath = filePath;
+    }
+
+    public ArrayList<Task> load() throws LunaBotException {
```
nigellqx: minor nitpick - perhaps there could be comments in this method to guide the readers?

nigellqx (commented):
LGTM - just some minor considerations regarding the code quality

PR #564, opened 2024-08-30
[Seah Jia Yi] iP

JiaYi-Gallium369:
# Gallium
> "Talk is cheap. Show me the code." ― Linus Torvalds

Gallium is more than your regular todo app. It's
- text-based
- ~~easy~~ SUPER EASY to use

All you need to do is,
1. Download the app [here] {https://github.com/JiaYi-Gallium369/ip}
2. Add your tasks
3. Let it manage your tasks! :shipit: 

And it is **FREE**!

Features:
- [x] Managing tasks
- [ ] Managing deadlines
- [ ] Reminders (coming soon)
- [ ] Managing events

Here's the `main` method:
```
        new Gallium("data/gallium.txt").run();
    }
```


src/main/java/MarkCommand.java
```diff
@@ -0,0 +1,25 @@
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class MarkCommand extends Command {
+    private String Message;
```
AngPengXuan: Should this be message instead to conform to coding standard? Variable names should be camel case.

src/main/java/Storage.java
```diff
@@ -0,0 +1,61 @@
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Scanner;
+
+public class Storage {
+    File filePath;
```
AngPengXuan: Should this be private? Same for the other 2 variables.

src/main/java/Parser.java
```diff
@@ -0,0 +1,89 @@
...
+
+    public Parser(Ui ui, TaskList taskList) {
+        this.ui = ui;
+        this.taskList = taskList;
+    }
+
+    public Command parse(String Message) {
+        while (!Message.equals(BYE)) {
+            try {
+                switch (Message) {
```
AngPengXuan: Should this switch statement be something like this instead:
```
switch (condition) {
case ABC:
    statements;
    // Fallthrough
case DEF:
    statements;
    break;
case XYZ:
    statements;
    break;
default:
    statements;
    break;
}
```

src/main/java/Ui.java
```diff
@@ -0,0 +1,118 @@
...
+        printAnyMessage("Please put the from and to of the event!!");
+    }
+
+    public void showWrongIndex() {
+        printAnyMessage("Please put a number between 1 and " + (Task.count - 1)
+                + "!" + "\n    Now you have " + (Task.count - 1) + " " + Task.taskCountDelete()
+                + " in the list.");
+    }
+
+    public void showIOException(IOException e) {
```
AngPengXuan: Should this be showIoException instead?

src/main/java/Event.java
```diff
@@ -0,0 +1,76 @@
...
+            this.fromDate = from.split(", ")[0];
+            this.fromTime = from.split(", ")[1];
+            this.toDate = to.split(", ")[0];
+            this.toTime = to.split(", ")[1];
+        }
+
+    }
+
+    public String toString() {
+        return "[E]" + this.getStatusIcon() + this.desc + " (from: " +
```
AngPengXuan: Should the '+' be on the next line instead?

src/main/java/Gallium.java
```diff
@@ -0,0 +1,39 @@
...
+    private TaskList taskList;
+    private Ui ui;
+    private Parser parser;
+
+    public Gallium(String filePath) {
+        ui = new Ui();
+        storage = new Storage(filePath);
+        // try {
+        taskList = new TaskList(storage.load());
+        // } catch (GalliumException e) {
```
AngPengXuan: Perhaps these comments can be removed? They should already be captured in your git history.

src/main/java/Event.java
```diff
@@ -0,0 +1,76 @@
...
+                this.toDate = LocalDate.parse(toDateString).format(DateTimeFormatter.ofPattern("MMM d yyyy"));
+                SimpleDateFormat inputTime = new SimpleDateFormat("HHmm");
+                SimpleDateFormat outputTime = new SimpleDateFormat("hh:mm a");
+                Date time24 = inputTime.parse(to.split(" ")[1]);
+                this.toTime = outputTime.format(time24);
+            } catch (ParseException e) {
+                e.printStackTrace();
+            }
+        } else if (description.startsWith("[E][ ] ")) {
+            String[] parts = description.split("\\[E\\]\\[ \\] ");
```
AngPengXuan: There are quite a lot of repeated code between this if and the one below, maybe can create variables to avoid repetition?

src/main/java/Deadline.java
```diff
@@ -0,0 +1,52 @@
...
+                this.date = LocalDate.parse(dateString).format(DateTimeFormatter.ofPattern("MMM d yyyy"));
+                SimpleDateFormat inputTime = new SimpleDateFormat("HHmm");
+                SimpleDateFormat outputTime = new SimpleDateFormat("hh:mm a");
+                Date time24 = inputTime.parse(by.split(" ")[1]);
+                this.time = outputTime.format(time24);
+            } catch (ParseException e) {
+                e.printStackTrace();
+            }
+        } else if (description.startsWith("[D][ ] ")) {
+            String[] parts = description.split("\\[D\\]\\[ \\] ");
```
AngPengXuan: Both this if segment and the one below looks very similar, maybe can create another function containing the similar parts so as not to repeat so much code?

src/main/java/AddCommand.java
```diff
@@ -0,0 +1,28 @@
+public class AddCommand extends Command {
+    private String desc;
+
+    public AddCommand(String desc) {
+        this.desc = desc;
+    }
+
+    @Override
+    public void execute(TaskList taskList, Ui ui, Storage storage) throws GalliumException {
```
AngPengXuan: Good job making use of OOP here to add tasks!

src/main/java/Parser.java
```diff
@@ -0,0 +1,89 @@
...
+                    case "todo":
+                    case "todo ":
+                    case "deadline":
+                    case "deadline ":
+                    case "event":
+                    case "event ":
+                        throw new GalliumException("OOPS!!! The description of a " + Message + " cannot be empty.");
+
+                    default:
+                        if (Message.matches(MARK + " \\d+") || Message.matches(UNMARK + " \\d+")) {
```
AngPengXuan: Maybe can add this to the cases above? Default probably shouldn't contain so many if else statements.

AngPengXuan (commented):
Overall, I found your code easy to read except the switch statement with nested if-else.  I also noted a few minor coding standard violations
too. Good job making use of OOP for Tasks and Commands!

PR #566, opened 2024-08-30
[sayomaki] iP

sayomaki:
# Alexer
> The future depends on what you do today - Mahatma Gandhi

A lexer that can do (**quite**) a few things. (is this a pun? maybe)
It takes care of:
- tracking your completed tasks
- watching your un-completed tasks
- ensuring you don't forget your tasks

Most importantly, it is:
- [x] *very* easy to use
- [x] simple to understand
- [ ] complicated to setup

So what are you waiting for? Download and try today! 
1. Clone the [repo](https://github.com/sayomaki/ip) or download the jar
2. Compile the code (if needed): `./gradlew shadowJar`
3. Run it and enjoy! 😄 

---

Also, here is a quick way to create multi-line text in Java! Check this out!
```java
String longLongText = """
     wee
     woo

     I am a multiline text, yipee!""";
```


src/main/java/Alexer.java
```diff
@@ -0,0 +1,224 @@
...
+
+public class Alexer {
+    private static final String GOODBYE = "Goodbye! If you ever want to chat again, I'll be here.\nHave a great day!";
+    private static final String BREAK = "____________________________________________________________";
+
+    public static final String NAME = "Alexer";
+
+    private final Scanner scanner = new Scanner(System.in);
+
+    private final TaskManager tasks = new TaskManager();
```
GabrielCWT: Maybe rename this to taskManager? I would think tasks would represent the tasks themselves e.g. an arraylist.

GabrielCWT (commented):
Not really much complaints about code standard. Keep up the good work!

src/main/java/Deadline.java
```diff
@@ -0,0 +1,37 @@
+public class Deadline extends Task {
+    protected String deadline;
+
+    public Deadline(String description, String deadline) {
+        super(description);
+        this.deadline = deadline;
+        type = TaskType.DEADLINE;
```
Green-Tea-123: Perhaps you can align with the above statement and put `this.type` instead of just `type`. This applies for the other task subclasses as well.

src/main/java/Todo.java
```diff
@@ -0,0 +1,34 @@
+public class Todo extends Task {
+    public Todo(String description) {
+        super(description);
+        type = TaskType.TODO;
+    }
+
+    public static Todo fromTaskString(String taskString) {
+        String[] parts = taskString.split("\\|");
+        if (parts.length < 3)
+            return null;
```
Green-Tea-123: Perhaps you should add open and close braces Egyptian style to comply with the coding standard.

Green-Tea-123 (commented):
Other than the comments I have made, your code complies with the coding standard for this course. Good job!

src/main/java/Alexer.java
sayomaki: Thanks for the suggestion, will take that into consideration!

src/main/java/Deadline.java
sayomaki: I was trying to follow the code guidelines for this! There is a clause on this: "Avoid unnecessary use of this with fields."

For `deadline` there is a conflicting name in the method parameter and as part of the object, so the explicit `this` keyword was necessary. However, for the `type` field there was no such ambiguity so I went ahead and omitted `this`!

src/main/java/Todo.java
sayomaki: Noted, thanks for pointing this out!

PR #575, opened 2024-08-30
[Quasant] iP

Quasant:
# Github got Screwllum before HSR 💯 

> He might come one day, but not today.

Ever wanted someone you can ask to help you remember the things you need to do? Someone that is ...
- Handsome
- Genius
- Metallic
- Costs the grand total of ~~1600 stellar jades~~ **free!**

All you need to do, is to 
1. Download the jar file from [here](https://github.com/Quasant/ip/releases/tag/v0.1)
2. Open the application
3. Type in the commands! 

Features:

- [x] Basic functionality
- [x] A beautiful GUI
- [ ] Can be used to clear pure-fiction (coming soon in 4.x)
  
---
Here is the `main` function.

```java 
public class Main {
    }
}
```


src/main/java/Alvis.java
```diff
@@ -0,0 +1,129 @@
...
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileWriter;
+import java.io.IOException;
+
+
+public class Alvis {
+    private static final String FILE_PATH = "./Save.txt";
+    
+    public static void main(String[] args) throws IOException {
```
thisisaditya17: Hmmm... 🤔 
Isn't main method too cramped right now?

Suggestion: Consider creating a different method for processing the input instead of putting everything inside main. It can cause issues when you'll be trying to further scale it and will make error handling and testing difficult. 😸 
Can create a method like startChatbot().

src/main/java/Alvis.java
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
+
+    public static void addTask(Task task, List<Task> taskList) {
+        taskList.add(task);
+        System.out.println("Understood, I have added to your list:\n" + task.toString());
+        System.out.println("You have " + taskList.size() + " tasks");
+    }
+
+    private static void loadTasks(List<Task> taskList) throws IOException {
```
thisisaditya17: Idk if you have reached to that Level in the ip Progress where you'll have to implement better OOP. 
But can separate saving and loading tasks into a different class. 🐱 🐈

src/main/java/Alvis.java
```diff
@@ -0,0 +1,129 @@
...
+                    int index = Integer.parseInt(userInput.substring(7)) - 1;
+                    taskList.remove(index);
+                    System.out.println("As requested, I have deleted the task");
+                } catch (IndexOutOfBoundsException e) {
+                    System.out.println("I cannot do that, the index is invalid");
+                }
+                writeToFile(taskList);
+                break;
+            case "deadline":
+                int byIndex = userInput.indexOf("/by");
```
thisisaditya17: I feel rn, the main method is doing everything. Must be tiring for it. 😩 

Suggestion: Instead of main method handling everything like taking input and processing it, you can break those tasks into different classes like ProcessTasks.class, SaveAndLoad.class, etc. according to your taste. 🕺🏼

thisisaditya17 (commented):
Overall, good code. Easy to understand. 
Just consider my sugesstions. 😄

src/main/java/Alvis/Deadline.java
```diff
@@ -0,0 +1,23 @@
+package Alvis;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+public class Deadline extends Task {
+    private LocalDateTime deadline;
```
Shantaa: Would it better to rename this variable, maybe something like "due"? So that others don't misunderstand this as an instance of the Deadline class

Shantaa (commented):
LGTM

src/main/java/Alvis/Deadline.java
Quasant: That is a good point! I had taken your suggestion and renamed it to `by`.

src/main/java/Alvis.java
Quasant: Hahaha indeed my bot was getting tired. I have split up the workload into various other classes. Hopefully, the code is still easy to understand!

PR #579, opened 2024-08-30
[Wen Xuan] iP

e1121208:
# Axel

> “You don't have to control your thoughts. You just have to stop letting them control you.” – Dan Millman

**Axel** frees your mind of having to remember things you need to do. It's:

- text-based
- easy to learn
- ~~fast~~ ~~SUPER FAST~~ **INSANELY FAST** to use

All you need to do is:

1. Download it from [here](https://github.com/e1121208/ip).
2. Double-click it.
3. Add your tasks.
4. Let it manage your tasks for you! :wink:

And it's **FREE**!

## Features:
- [x] Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming soon)

If you're a Java programmer, you can use it to practice Java too! Here's an example of the `main` method:

```java
public class Main {
    }
}

### CODE REVIEW

File: src/main/java/axel/Command.java [hunk 1]
```diff
@@ -0,0 +1,190 @@
...
+
+import java.io.IOException;
+import java.util.List;
+
+/**
+ * Represents a command that can be executed by the application.
+ * Each command manipulates the task list, interacts with the user interface,
+ * and may involve saving changes to storage.
+ */
+public interface Command {
```
Dinoman44, line 11: Nice naming convention and abstraction! But it may be better to place these classes as separate `.java` files and package them into a `command` package? This way you compartmentalize your code and its easier to debug.

src/main/java/axel/DeadlineTask.java
```diff
@@ -0,0 +1,52 @@
...
+import java.time.format.DateTimeFormatter;
+import java.util.Locale;
+
+/**
+ * Represents a task with a deadline.
+ * Inherits from {@link Task}.
+ */
+public class DeadlineTask extends Task {
+    /** The deadline for the task. */
+    protected LocalDateTime by;
```
Dinoman44, line 13: The name kind of makes sense, but maybe you could instead have the name be `deadline` or `endTime`?

src/main/java/axel/EventTask.java
```diff
@@ -0,0 +1,58 @@
...
+import java.time.format.DateTimeFormatter;
+import java.util.Locale;
+
+/**
+ * Represents a task with a start and end time.
+ * Inherits from {@link Task}.
+ */
+public class EventTask extends Task {
+    /** The start time of the event. */
+    protected LocalDateTime from;
```
Dinoman44, line 13: Same as before, you could probably name this `fromTime` or `startTime`.

src/main/java/axel/EventTask.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /** The end time of the event. */
+    protected LocalDateTime to;
```
Dinoman44, line 16: Same as before, this could be `endTime`.

src/main/java/axel/Storage.java
```diff
@@ -0,0 +1,98 @@
+package axel;
+
+import java.io.*;
```
Dinoman44, line 3: We can't have wildcard imports unfortunately 😭

src/main/java/axel/TaskList.java
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * Returns the number of tasks in the task list.
+     *
+     * @return The number of tasks
+     */
+    public int size() {
+        return tasks.size();
+    }
+
+    public List<Task> findTasksWithKeyword(String keyword) {
```
Dinoman44, line 87: CS2030S coming in handy I see. Unfortunately you forgot JavaDoc comment -4 for style 😆

src/main/java/axel/Axel.java
```diff
@@ -0,0 +1,80 @@
+package axel;
```
Dinoman44, line 1: I think its a good idea to have related classes in different packages e.g. a `task` package for task classes, a `command` package for commands, an `axelException` package for your `AxelException`s etc.

Dinoman44 (commented):
Pretty good job, looking forward to ~~copying~~ seeing your UI! Just some minor naming changes to make.

PR #581, opened 2024-08-30
[Yang Qifeng] iP

YangQF2002:
# Brock 
> "Big papa brock" - Me [source](https://www.youtube.com/watch?v=dQw4w9WgXcQ)

Brock frees your mind from having to remember things you need to do. It's
- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use 

All you need to do is, 
1. download it from [here](https://www.youtube.com/watch?v=dQw4w9WgXcQ). 
2. double-click it. 
3. add your tasks. 

And it is **FREE**!

Features: 
- [x] Managing tasks
- [ ] Reminders (coming soon) 

If you are a Java programmer, you can use it to practice Java too! Here's the `main` method! 
```
     /**
     * Main method, when called, runs the main logic of the chatbot.
     * @param args The command line arguments.
     **/
        new Brock().run();
    }
```


src/main/java/commands/DeadlineCommand.java
```diff
@@ -0,0 +1,115 @@
...
+                    dateTimeValues[0]);
+        } else {
+            return new Deadline(description.toString(),
+                    dateTimeValues[0],
+                    dateTimeValues[1]);
+        }
+    }
+
+    /**
+     * {@inheritDoc}
```
Andrew22Teoh, line 126: Good use of this JavaDoc parameter! Wasn't aware of it myself 😮

src/main/java/commands/EventCommand.java
```diff
@@ -0,0 +1,141 @@
...
+        tasks.addToList(eventTask);
+        ui.displayResponse("Got it. I've added this task:\n"
+                + "  "
+                + tasks.getTaskDetails(eventTask)
+                + '\n'
+                + tasks.getTasksSummary());
+
+        // Update the save file
+        storage.writeToFile(tasks.numTasks()
+                        + ". "
```
Andrew22Teoh: I think the auto-indentations on these lines may not be correct, would be good to fix! 😄 
I noticed a similar problem in your TodoCommand.java around lines 71-74, so be sure to check for those.

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,245 @@
...
+    private static final File SAVE_FILE = new File(FILE_PATH);
+
+    /**
+     * Removes closing bracket from the body string of each task.
+     * So that the correct date or time can be extracted for deadline and event tasks.
+     *
+     * @param target String fragment with closing bracket to be removed.
+     * @return String without the closing bracket.
+     */
+    private String removeCloseBracket(String target) {
```
Andrew22Teoh: Would it be productive to have a check here to verify that the last character in target is indeed a close bracket, just to account for possible cases of corrupted data?

Andrew22Teoh (commented):
Very good! Coding standard is very strictly adhered to with a few minor deviations, and OOP structure is impressive 👍

src/main/java/commands/EventCommand.java
YangQF2002: Based on the style guide it says indentation of wrapped lines is 8 spaces (ie: 2x the usual 4 spaces), so i think this is correct? 😅

src/main/java/storage/Storage.java
YangQF2002: Agreed, good idea to include checks for save file corruption in general throughout the code 💯

src/main/java/commands/EventCommand.java
Andrew22Teoh: Oh, it shows as 16 spaces on my end...but as long as it looks right on your machine, all good!
image

YangQF2002: EH my bad i was looking at the wrong part, yea i fixed it now, thankss 👍

PR #598, opened 2024-08-31
[shotnothing] iP

shotnothing:
# Bob
“Your mind is for having ideas, not holding them.” – David Allen (source)

Bob frees your mind of having to remember things you need to do. It's:

- Text-based  
- Easy to learn  
- **FAST** SUPER **FAST** to use  

All you need to do is:

1. [Download it from here](#).
2. Double-click it.
3. Add your tasks.
4. Let it manage your tasks for you 😉  

And it is **FREE!**

## Features:
- [x] Managing tasks  
- [ ] Reminders (coming soon)

```java
public class Main {
    }
}

### CODE REVIEW

File: src/main/java/Bob/Parser.java [hunk 1]
```diff
@@ -0,0 +1,125 @@
...
+            throw new BobException("The date and time format is incorrect: " + e.getMessage());
+        }
+    }
+
+    /**
+     * Command to list all tasks.
+     *
+     * @param tasks
+     */
+    public static void processListCommand(List<Task> tasks) {
```
XinghuaJulia: processListCommand can be rather vague, do you mean execute?

src/main/java/Bob/Ui.java
```diff
@@ -0,0 +1,53 @@
...
+import bob.task.Task;
+
+public class Ui {
+
+    /**
+     * Prints a dialogue.
+     *
+     * @param input
+     */
+    public static void dialogue(String input) {
```
XinghuaJulia: dialogue is really vague, what kind of dialogue? Perhaps dialogueFormat would be better?

src/main/java/Bob/Bob.java
```diff
@@ -0,0 +1,18 @@
+package bob;
+
+import bob.task.Task;
+
+import java.util.ArrayList; 
+import java.util.List;
+import java.util.Scanner;
+
+public class Bob {
```
XinghuaJulia, line 12: Please leave header comments for what the class does.

src/main/java/Bob/Parser.java
```diff
@@ -0,0 +1,125 @@
...
+
+    /**
+     * Command to mark, unmark, or delete a task.
+     *
+     * @param userInput
+     * @param tasks
+     * @throws BobException
+     * @throws IOException
+     */
+    public static void processTaskModificationCommands(String userInput, List<Task> tasks) throws BobException, IOException {
```
XinghuaJulia: processTaskModificationCommand, while explained in the header, is rather confusing and unintuitive... Perhaps you might want to consider creating separate static methods or even classes for them?

XinghuaJulia (commented):
Looks good to me, just a few minor nits... Otherwise, good job!!!

src/main/java/Bob/Parser.java
```diff
@@ -0,0 +1,125 @@
...
+import bob.task.Todo;
+
+class Parser {
+
+    /**
+     * Parses a string into a LocalDateTime object.
+     *
+     * @param input
+     * @return LocalDateTime
+     * @throws BobException
```
rollingpencil, line 26: Please add a description of the parameters. I noticed the same issue in several other places below too.

src/main/java/Bob/BobException.java
```diff
@@ -0,0 +1,8 @@
+package bob;
+
+class BobException extends Exception {
+
+    public BobException(String message) {
```
rollingpencil: Please include JavaDoc comments on the purpose of the exception class for the class header comment and the constructor.

src/main/java/Bob/Parser.java
rollingpencil: This line exceeds 120 characters. Perhaps it would be better to warp it to the next line for better readability? I noticed the same issue in several method headers and individual lines below too.

src/main/java/Bob/Bob.java
```diff
@@ -0,0 +1,18 @@
+package bob;
+
+import bob.task.Task;
+
+import java.util.ArrayList; 
+import java.util.List;
+import java.util.Scanner;
```
rollingpencil:
```suggestion
import java.util.ArrayList; 
import java.util.List;
import java.util.Scanner;

import bob.task.Task;
```

The order of imports in this file seem to differ from the rest. Typicall you did
```java
import java.____

import bob.____
```
Perhaps you can change the import order to be consistent with the rest of the files?

src/main/java/Bob/Storage.java
```diff
@@ -0,0 +1,55 @@
...
+
+import bob.task.Task;
+
+class Storage {
+
+    /**
+     * Saves the tasks to a file.
+     *
+     * @param tasks
+     * @throws IOException
```
rollingpencil, line 19: Please add a description of the parameters. I noticed the same issue in several other places below too.

src/main/java/Bob/Ui.java
rollingpencil: I agree with Julia that `dialogue` as a variable name is vague. In this context, maybe `printDialogue` might be better?

src/main/java/Bob/Ui.java
```diff
@@ -0,0 +1,53 @@
...
+                String userInput = scanner.nextLine().trim();
+
+                if (userInput.equalsIgnoreCase("bye")) {
+                    Ui.dialogue("Bye. Hope to see you again soon!");
+                    break;
+                } else if (userInput.equalsIgnoreCase("list")) {
+                    Parser.processListCommand(tasks);
+                } else if (userInput.startsWith("mark") || userInput.startsWith("unmark") || userInput.startsWith("delete")) {
+                    Parser.processTaskModificationCommands(userInput, tasks);
+                } else if (userInput.startsWith("todo") || userInput.startsWith("deadline") || userInput.startsWith("event")) {
```
rollingpencil:
```suggestion
                } else if (userInput.startsWith("mark") || userInput.startsWith("unmark") 
                        || userInput.startsWith("delete")) {
                    Parser.processTaskModificationCommands(userInput, tasks);
                } else if (userInput.startsWith("todo") || userInput.startsWith("deadline") 
                        || userInput.startsWith("event")) {
```

This line exceeds 120 characters. Perhaps it would be better to warp it to the next line for better readability?

src/main/java/Bob/Parser.java
```diff
@@ -0,0 +1,125 @@
...
+import java.time.format.DateTimeParseException;
+import java.util.List;
+import java.util.Locale;
+
+import bob.task.Deadline;
+import bob.task.Event;
+import bob.task.Task;
+import bob.task.Todo;
+
+class Parser {
```
rollingpencil, line 19: Please include JavaDoc comments on the purpose of the Parser class.

src/main/java/Bob/task/Task.java
```diff
@@ -0,0 +1,49 @@
+package bob.task;
+
+public class Task {
```
rollingpencil, line 3: Please include JavaDoc comments on the purpose of the Task class as well as in the subclasses inherited from this class (Todo, Event and Deadline).

rollingpencil (commented):
Good job! Just a few nits to pick here.
1. Several classes and their methods either have missing JavaDoc comments or the description of the parameters is missing.
2. There are several lines in some of the files where you have exceeded the hard 120 character limit.

I have only highlighted some of the examples to remove clutter in the review, but it would be helpful to take a look at the entire codebase as a whole to see where this problem may have occurred

src/main/java/Bob/MainWindow.java
```diff
@@ -0,0 +1,111 @@
...
+
+        sendButton.setPrefWidth(55.0);
+
+        AnchorPane.setTopAnchor(scrollPane, 1.0);
+
+        AnchorPane.setBottomAnchor(sendButton, 1.0);
+        AnchorPane.setRightAnchor(sendButton, 1.0);
+
+        AnchorPane.setLeftAnchor(userInput, 1.0);
+        AnchorPane.setBottomAnchor(userInput, 1.0);
```
joeltio, line 100: Consider grouping these statements by the target component (i.e. do everything related to `sendButton` first then `userInput` instead of doing all the `AnchorPane` related items then `prefWidth` properties)

This is important because usually you'll look for things by component, not by their attribute

src/main/java/Bob/Parser.java
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     * Command to list all tasks.
+     *
+     * @param tasks
+     */
+    public static String executeListCommand(List<Task> tasks) {
+        String out = "Here are the tasks in your list: \n";
+        for (int i = 0; i < tasks.size(); i++) {
+            out += (i + 1) + ". " + tasks.get(i) + "\n";
+        }
```
joeltio, line 47: Consider using `StringBuilder` here as it is more performant.

src/main/java/Bob/Parser.java
```diff
@@ -0,0 +1,137 @@
...
+     */
+    public static LocalDateTime parseDateTime(String input) throws BobException {
+        try {
+            // Correct format: "dd/MM/yyyy HHmm" (e.g., "02/12/2019 1800")
+            DateTimeFormatter formatter = DateTimeFormatter.ofPattern("d/M/yyyy HHmm", Locale.ENGLISH);
+            return LocalDateTime.parse(input, formatter);
+        } catch (DateTimeParseException e) {
+            throw new BobException("The date and time format is incorrect: " + e.getMessage());
+        }
+    }
```
joeltio, line 36: This method seems out of place consider the purpose of `Parser`. Perhaps if it is not used anywhere else you can make it private? It is more of a helper function to `Parser` as compared to something that `Parser` does that other classes can use. The main thing `Parser` seems to do is to interpret and execute commands, not to parse `DateTime`s.

Alternatively, move it to another class

src/main/java/Bob/Parser.java
```diff
@@ -0,0 +1,137 @@
...
+     * @param userInput
+     * @param tasks
+     * @throws BobException
+     * @throws IOException
+     */
+    public static String processTaskCreationCommands(String userInput, List<Task> tasks) 
+            throws BobException, IOException {
+        Task newTask = null;
+        if (userInput.startsWith("todo")) {
+            String description = userInput.substring(5).trim();
```
joeltio: I would suggest that this be refactored to another method, especially because of the calculation of `5`. If, in the future, `todo` is changed to `todos`, there is a chance that `5` is not changed to `6` as well.

src/main/java/Bob/Parser.java
```diff
@@ -0,0 +1,137 @@
...
+     * Creates a new task based on the user input and adds it to the list of
+     * tasks.
+     *
+     * @param userInput
+     * @param tasks
+     * @throws BobException
+     * @throws IOException
+     */
+    public static String processTaskCreationCommands(String userInput, List<Task> tasks) 
+            throws BobException, IOException {
```
joeltio, line 150: Perhaps you may want to hide the logic of how a task is stored to within this method?
Currently, if this method throws `IOException`, all other usages of this method will need to know that this method performs some sort of IO. If, in the future, the IO is removed, then all the usages will need to be updated to not have try catches.

Instead, it may be better to catch the `IOException` and rethrow as a `BobException`

src/main/java/Bob/Ui.java
```diff
@@ -0,0 +1,56 @@
...
+ * Represents the CLI user interface of Bob, depreciated for the GUI. Will be removed in future versions.
+ */
+public class Ui {
+
+    /**
+     * Prints a dialogue.
+     *
+     * @param input
+     */
+    public static void printDialogueCLI(String input) {
```
joeltio, line 19:
```suggestion
    public static void printDialogueCli(String input) {
```
Should be `Cli` instead of `CLI` following code conventions

src/main/java/Bob/task/Deadline.java
```diff
@@ -0,0 +1,43 @@
...
+     * Converts a LocalDateTime object to a string.
+     *
+     * @param dateTime
+     * @return String
+     */
+    private static String dateTimeToString(LocalDateTime dateTime) {
+        // Correct format: "MMM dd yyyy" e.g., (Oct 15 2019)
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("MMM dd yyyy HH:mm", Locale.ENGLISH);
+        return dateTime.format(formatter);
+    }
```
joeltio: Currently, this method is also defined in `Event`, which is fine. But personally I foresee that this would be reused quite a bit. Is there any reason why you wouldn't want to refactor this (together with the complement function in `Parser`) to another file?

src/main/java/Bob/task/Todo.java
```diff
@@ -0,0 +1,13 @@
+package bob.task;
+
+public class Todo extends Task {
+
+    public Todo(String description) {
+        super(description);
+    }
```
joeltio, line 11: I don't think this constructor is needed, is there a reason why it is implemented here?

PR #599, opened 2024-08-31
rynlwhh iP

rynlwhh:
# NaegaBot

> “Simplicity is the soul of efficiency.” — *Austin Freeman*

**NaegaBot** is here to help you manage tasks efficiently and effortlessly! It’s:

- Text-based  
- Easy to learn  
- ~~Overwhelming~~ **Super fast** to use ⚡

To get started:

1. [Download it here](https://github.com/rynlwhh/ip).
2. Double-click the app to launch it.
3. Start managing your tasks with simple commands.
4. Let NaegaBot do the rest! 😄

### Features:

- Managing tasks
- Managing deadlines
- Reminders (coming soon)

If you’re a Java developer, you can use it to practice your Java skills. Here’s the main method:

```java
public class Main {
    }
}

### CODE REVIEW

File: src/main/java/Event.java [hunk 1]
```diff
@@ -0,0 +1,15 @@
+public class Event extends Task {
+    protected String from;
+    protected String to;
```
RezwanAhmed123: I like the simplicity of the names of the time variables. However, it may be abit more clearer if the variables were more specific such as the following:

```suggestion
    protected String fromDateTime;
    protected String toDateTime;
```

src/main/java/Naega.java
```diff
@@ -0,0 +1,150 @@
...
+        System.out.println("____________________________________________________________");
+    }
+
+    private static void handleUnmark(ArrayList<Task> tasks, String userInput) throws NaegaException {
+        int taskIndex = parseTaskIndex(userInput, "unmark ");
+        tasks.get(taskIndex).markAsNotDone();
+        System.out.println("____________________________________________________________");
+        System.out.println(" OK, I've marked this task as not done yet:");
+        System.out.println("   " + tasks.get(taskIndex));
+        System.out.println("____________________________________________________________");
```
RezwanAhmed123: I like how these methods were named. Good job!

Perhaps common code could be extracted out tho.

src/main/java/Naega.java
```diff
@@ -0,0 +1,150 @@
...
+        }
+        String description = parts[0];
+        String[] timeParts = parts[1].split(" /to ");
+        if (timeParts.length < 2) {
+            throw new NaegaException("The event must include a '/to' keyword.");
+        }
+        Task newTask = new Event(description, timeParts[0], timeParts[1]);
+        tasks.add(newTask);
+        printTaskAdded(tasks, newTask);
+    }
```
RezwanAhmed123: Similar to the previous comment, perhaps, the common code could be extracted into a separate method?

src/main/java/Naega.java
```diff
@@ -0,0 +1,150 @@
...
+                }
+            } catch (NaegaException e) {
+                System.out.println("____________________________________________________________");
+                System.out.println(" Oops! " + e.getMessage());
+                System.out.println("____________________________________________________________");
+            } catch (NumberFormatException | IndexOutOfBoundsException e) {
+                System.out.println("____________________________________________________________");
+                System.out.println(" Oops! Invalid task number.");
+                System.out.println("____________________________________________________________");
+            }
```
RezwanAhmed123: I like how mostly clean this block of code is! Any reason for using if-else instead of switch-case here though?

RezwanAhmed123 (commented):
LGTM, very readable except some nitpicky comments. Some repetition in code could be removed by extracting common code in multiple methods. Overall well done!

PR #625, opened 2024-09-02
[Nicholascyx] ip

Nicholascyx:
# Kafka
> "The path to the future begins right here." - Kafka

Kafka aids you in following the path you are destined to be. It's,
- text-based
- easy to learn 
- ~~FUN~~ *SUPER* FUN to use

All you need to do is, 
1. download it from [here](https://github.com/Nicholascyx/ip/releases/tag/A-Release). 
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you 👍 

And it is **FREE**!

Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)

Here's the `main` method:
```ruby
public class KafkaLauncher {
    }
}
```


src/main/java/Kafka.java
```diff
@@ -0,0 +1,136 @@
...
+                } else if (userInput[0].equalsIgnoreCase("unmark")) {
+                    int taskNumber = Integer.parseInt(userInput[1]);
+                    kafka.unmark(taskNumber);
+                } else if (userInput[0].equalsIgnoreCase("delete")) {
+                    int taskNumber = Integer.parseInt(userInput[1]);
+                    kafka.delete(taskNumber);
+                } else {
+                    if (userInput[0].equalsIgnoreCase("todo")) {
+                        if (userInput.length < 2) {
+                            throw new KafkaException("It seems you've left the details blank. Even the simplest tasks need some direction, don't you think?");
```
hm-yang1: Exception message is quite long and used repeatedly. Maybe store the message as a String variable? Or write a function to throw a new exception?

src/main/java/Event.java
```diff
@@ -0,0 +1,16 @@
+public class Event extends Task {
+
+    protected String from;
```
hm-yang1: I don't see the protected access modifier often. Quite curious about the reason

src/main/java/Kafka.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    public static void main(String[] args) {
+        String logo = """
+                   __  __            __     _
+                  |  |/  /  ____   _/  /_  | |      ____
+                  |     /  / _  | |_    _| | |/ /  / _  |
+                  |     \\ | |_| |   |  |   |   <  | |_| |
+                  |__|\\__\\ \\____|   |__|   |_|\\ \\  \\____|
+                """;
```
hm-yang1: Nice. I wouldn't be able to create that and adds character to the bot

src/main/java/Kafka.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    public void addTask(Task task) {
+        this.tasks.add(task);
+        System.out.println("  Got it. I've added this task.");
+        System.out.println("    " + task);
+        System.out.println("  Now you have " + this.tasks.size() + " task(s) in the list.");
+    }
+
+    public void createList() {
```
hm-yang1: The function name is a little misleading as it does not create a new list. Perhaps getList or printList might be better?

src/main/java/Kafka.java
```diff
@@ -0,0 +1,136 @@
...
+                    int taskNumber = Integer.parseInt(userInput[1]);
+                    kafka.mark(taskNumber);
+                } else if (userInput[0].equalsIgnoreCase("unmark")) {
+                    int taskNumber = Integer.parseInt(userInput[1]);
+                    kafka.unmark(taskNumber);
+                } else if (userInput[0].equalsIgnoreCase("delete")) {
+                    int taskNumber = Integer.parseInt(userInput[1]);
+                    kafka.delete(taskNumber);
+                } else {
+                    if (userInput[0].equalsIgnoreCase("todo")) {
```
hm-yang1: Nested if statements are not as easy to read. Is it possible to just add more elif to handle the add task cases instead of nesting them in the else block?

hm-yang1 (commented):
Looks good.
Requested some cosmetic changes.

src/main/java/kafka/Storage.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    public ArrayList<Task> load() throws FileNotFoundException {
+        this.getNewFile(filePath);
+        ArrayList<Task> tasks = new ArrayList<>();
+        this.printFileContents(tasks);
+        return tasks;
+    }
+
+    public void printFileContents(ArrayList<Task> tasks) {
```
blackpanther9229: You definitely need JavaDoc comment for this method

src/main/java/kafka/Storage.java
```diff
@@ -0,0 +1,89 @@
...
+                            + event.description + " | " + event.from + " | " + event.to
+                            + System.lineSeparator());
+                }
+            }
+        } catch (IOException e) {
+            System.out.println("Something went wrong: " + e.getMessage());
+        }
+    }
+
+    public void getNewFile(String filePath) {
```
blackpanther9229: This method should have JavaDoc comment

src/main/java/kafka/Parser.java
```diff
@@ -0,0 +1,93 @@
+package kafka;
+
+import java.io.IOException;
+import java.time.format.DateTimeParseException;
+import java.time.LocalDateTime;
+
+public class Parser {
+
+    public static void parseCommand(String userInput, TaskList taskList, Storage storage, Ui ui) throws KafkaException, IOException, DateTimeParseException {
```
blackpanther9229: you need JavaDoc comment for this

src/main/java/kafka/Parser.java
```diff
@@ -0,0 +1,93 @@
...
+                if (eventParts.length < 3) {
+                    throw new KafkaException("It appears the details for this event task are off. Let's give it another go, shall we?");
+                }
+                LocalDateTime from = LocalDateTimeConverter.getLocalDateTime(eventParts[1]);
+                LocalDateTime to = LocalDateTimeConverter.getLocalDateTime(eventParts[2]);
+                Task event = new Event(eventParts[0], from, to, false);
+                taskList.addTask(event);
+                storage.writeToFile(taskList.tasks);
+                ui.addTask(event, taskList);
+                break;
```
blackpanther9229: It is better to add some comments for clarifying

src/main/java/kafka/TaskList.java
```diff
@@ -0,0 +1,51 @@
...
+
+    public void addTask(Task task) {
+        this.tasks.add(task);
+    }
+
+    public void printList() {
+        if (tasks.isEmpty()) {
+            return;
+        }
+        String listMessage = "";
```
Oceankoh, line 46: Minor nitpick, but you could consider using StringBuilder for some improvement in time complexity. It is really minor though as the strings are probably not long enough to result in any real world performance benefits :)

src/main/java/kafka/Parser.java
```diff
@@ -0,0 +1,93 @@
...
+        String[] splitInput = userInput.trim().split(" ", 2);
+
+        if (splitInput[0] == null) {
+            return;
+        }
+
+        String command = splitInput[0].toLowerCase();
+        String arguments = splitInput.length > 1 ? splitInput[1] : "";
+
+        switch (command) {
```
Oceankoh: Wow! This is a really long switch case. Could this be broken up into smaller parts? Maybe try creating a method for each case! This would improve code readability!

src/test/java/seedu/kafka/StorageTest.java
```diff
@@ -0,0 +1,65 @@
...
+import kafka.Todo;
+import kafka.Event;
+import kafka.Task;
+import org.junit.jupiter.api.Test;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.time.LocalDateTime;
+import java.util.ArrayList;
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
Oceankoh, line 14: Does this follow the package import coding standards?

src/main/java/kafka/Kafka.java
```diff
@@ -0,0 +1,51 @@
...
+                ui.showError(e);
+            } catch (DateTimeParseException e) {
+                ui.incorrectDateDetails();
+            }
+        }
+        ui.goodbye();
+    }
+
+    public static void main(String[] args) {
+        new Kafka("C:/Users/Nicholas/Downloads/Kafka.txt").run();
```
Oceankoh: Would this work if I run it on a non-windows PC?

Oceankoh (commented):
Great Effort! Loved how you customized the different print statements

src/test/java/seedu/kafka/StorageTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+
+public class StorageTest {
+
+    @Test
+    public void testWriteTodoToFile() throws IOException {
```
Elinengu, line 20: I think this might not be the naming convention for testing method. Perhaps can double check it on topics?

src/main/java/kafka/Kafka.java
Elinengu: Hello, I think it might not work on other computer (even if it's windows) because the path directory won't be the same. Perhaps it's better to use a relative path instead? And maybe it's better to handle the case where a folder or a txt file has not being created yet to prevent checked exception?
Screenshot 2024-09-07 191711

src/main/java/kafka/Storage.java
```diff
@@ -0,0 +1,89 @@
...
+                            + todo.description + System.lineSeparator());
+                } else if (t instanceof Deadline deadline) {
+                    fw.write("D | " + deadline.isDone + " | "
+                            + deadline.description + " | " + deadline.by + System.lineSeparator());
+                } else if (t instanceof Event event) {
+                    fw.write("E | " + event.isDone + " | "
+                            + event.description + " | " + event.from + " | " + event.to
+                            + System.lineSeparator());
+                }
+            }
```
Elinengu, line 99: I'm not quite sure about it (since I couldn't run the file) but I think this method might cause only 1 line of task to be written in the final txt file because the writer object is not being created in append mode. Perhaps it's better to check it out?

PR #673, opened 2024-09-07
[RobotWizzard] iP

RobotWizzard:


src/main/java/Bob.java
```diff
@@ -0,0 +1,207 @@
...
+        | |              | | |              | | |              | |
+        | '--------------' | '--------------' | '--------------' |
+        '----------------' '----------------' '----------------'
+        """;
+        Scanner scanner = new Scanner(System.in);
+
+        System.out.println(logo);
+        greet();
+        while (true) {
+            boolean executed = false;
```
LowXiSi: Should executed be named as isExecuted instead to show that it is a boolean value?

src/main/java/Bob.java
```diff
@@ -0,0 +1,207 @@
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Scanner;
+
+public class Bob {
+    private static final String SEPARATOR = "____________________________________________________________";
+    private static final String LINE_PREFIX = "    ";
+
+    private static String argument = "";
+    private static final List<Task> list = new ArrayList<>();
```
LowXiSi: Should list be named as tasks instead to represent a collection of tasks?

LowXiSi (commented):
Overall, most of your code seems to follow the coding standard! :)

src/main/java/Bob.java
```diff
@@ -0,0 +1,207 @@
...
+        LIST("list") {
+            @Override
+            public void run() {
+                if (list.isEmpty()) {
+                    say("You have not added any tasks yet.");
+                } else {
+                    StringBuilder text = new StringBuilder();
+                    int i;
+                    for (i = 0; i < list.size() - 1; ++i) {
+                        text.append(i + 1).append(".").append(list.get(i)).append("\n");
```
potaotototo: Is there anyway that this could be abstracted? For example, maybe a method inside the Task class or its inheriting classes?

src/main/java/Bob.java
```diff
@@ -0,0 +1,207 @@
...
+                    Task task = getTask(fromIndex, toIndex);
+                    list.add(task);
+                    say("added: " + task);
+                }
+            }
+
+            private static Task getTask(int fromIndex, int toIndex) {
+                String desc = fromIndex < toIndex
+                                ? argument.substring(0, fromIndex)
+                                : argument.substring(0, toIndex);
```
potaotototo: Not sure whether this is accurate? Maybe try to add comments to explain why you choose to incorporate fromIndex and toIndex?

src/main/java/Bob.java
```diff
@@ -0,0 +1,207 @@
...
+        },
+        MARK("mark") {
+            @Override
+            public void run() {
+                if (argument.isBlank()) {
+                    throw new MissingArgumentException("index of the task that you want to mark");
+                }
+                int i;
+                try {
+                    i = Integer.parseInt(argument) - 1;
```
potaotototo: Not compulsory but could try to name i as something else. This can help other programmers understand the code easily.

src/main/java/Bob.java
```diff
@@ -0,0 +1,207 @@
...
+            }
+        },
+        CATCH_ALL("") {
+            @Override
+            public void run() {
+                throw new UnknownCommandException();
+            }
+        };
+
+        public final String CMD;
```
potaotototo: I think the coding standards mentioned that we should avoid short forms? Maybe can name as COMMAND instead of CMD?

PR #760, opened 2024-09-12
 BCD-Extensions (D-Loans)

M-kew:
Implemented Loan Feature:
Created a new Loan class to handle loan-related operations in the Nuffle chatbot.
Added LoanCommand to facilitate user interaction with the loan feature, allowing users to initiate and manage loans within the chatbot.

Parser Modification:
Updated the Parser class to recognize and process loan-related user inputs.
Ensured that loan commands are parsed correctly and routed to the appropriate logic for execution.


M-kew:
LGTM.

PR #761, opened 2024-09-12
Bcd extension update

SSirMentos:
Deleted the ActionTaskTuple class as it was no longer being used in the codebase.


WeiHanWong:
Good job

